PR #2, opened 2021-06-01
Implement Gnosis Auction

chuddster:


contracts/vaults/RibbonThetaVault.sol
```diff
@@ -45,6 +56,13 @@ contract RibbonThetaVault is DSMath, GnosisAuction, OptionsVaultStorage {
     // https://github.com/opynfinance/GammaProtocol/blob/master/contracts/MarginPool.sol
     address public immutable MARGIN_POOL;
 
+    // Our library for all Opyn related logic
+    address public immutable GAMMA_PROTOCOL;
```
kenchangh, line 60: We don't actually need to store a reference to the GAMMA_PROTOCOL because it's already implicitly done by linking the library. We can just do `GammaProtocol.createShort` directly too I think

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -12,16 +12,27 @@ import {GnosisAuction} from "../protocols/GnosisAuction.sol";
...
-
-contract RibbonThetaVault is DSMath, GnosisAuction, OptionsVaultStorage {
+import {IGnosisAuction} from "../interfaces/IGnosisAuction.sol";
+import {
+    IStrikeSelection,
+    IOptionsPremiumPricer
+} from "../interfaces/IRibbon.sol";
+
+contract RibbonThetaVault is DSMath, OptionsVaultStorage {
+    using GammaProtocol for address;
```
kenchangh, line 22: We can remove this, see: https://github.com/ribbon-finance/ribbon-v2/pull/2/files#r646754461

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -205,6 +257,21 @@ contract RibbonThetaVault is DSMath, GnosisAuction, OptionsVaultStorage {
...
 
+    /**
+     * @notice Sets the new discount on premiums for options we are selling
+     * @param newPremiumDiscount is the premium discount
+     */
+    function setPremiumDiscount(uint256 newPremiumDiscount)
+        external
+        onlyManager
+    {
+        require(newPremiumDiscount > 0, "newPremiumDiscount != 0");
```
kenchangh, line 268: We should also set an upper bound for the discount to something reasonable like 30%

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -459,16 +526,53 @@ contract RibbonThetaVault is DSMath, GnosisAuction, OptionsVaultStorage {
...
 
+        startAuction();
+
         emit OpenShort(newOption, shortAmount, msg.sender);
     }
 
+    /**
+     * @notice Initiate the gnosis auction.
+     */
+    function startAuction() public onlyManager {
```
kenchangh, line 544: Is there a reason this function needs to be `public`?

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -459,16 +526,53 @@ contract RibbonThetaVault is DSMath, GnosisAuction, OptionsVaultStorage {
...
+            GNOSIS_EASY_AUCTION,
+            optionsPremiumPricer,
+            auctionDetails
+        );
+    }
+
+    /**
+     * @notice Burn the remaining oTokens left over from gnosis auction.
+     */
+    function burnRemainingOTokens() external onlyManager nonReentrant {
```
kenchangh, line 564: This process should be bundled together with closing the auction imo, so it would be like:

```solidity
function closeAuction() {
  GnosisAuction.settleAuction()
  burnRemainingOTokens()
}
```

kenchangh: I'm okay with the change being a follow-up PR because this PR is getting huge

migrations/ribbon_eth_covered_call_v2.js
```diff
@@ -0,0 +1,87 @@
+const RibbonThetaVault = artifacts.require("RibbonThetaVault");
```
kenchangh, line 1: We're no longer using Truffle anymore so we can't do `artifacts.require`, and do migrations.
It's more vanilla and we have to manually deploy with typescript files now

contracts/protocols/GnosisAuction.sol
```diff
@@ -1,4 +1,114 @@
...
+        IOtoken newOToken = IOtoken(oTokenAddress);
+        IGnosisAuction auction = IGnosisAuction(gnosisEasyAuction);
+        // We take our current oToken balance and we subtract an
+        // amount that is the fee gnosis takes. That will be our sell amount
+        // but gnosis will transfer all the otokens
+        oTokenSellAmount = uint96(
+            IERC20(oTokenAddress)
+                .balanceOf(address(this))
+                .mul(auction.FEE_DENOMINATOR())
+                .div(auction.FEE_DENOMINATOR().add(auction.feeNumerator()))
```
kenchangh, line 91: We need to check that this uint256 does not overflow uint96 (`type(uint96).max`)

contracts/protocols/GnosisAuction.sol
```diff
@@ -1,4 +1,114 @@
...
+                auctionDetails.underlying,
+                optionsPremiumPricer,
+                auctionDetails.premiumDiscount
+            );
+
+        IERC20(auctionDetails.oTokenAddress).safeApprove(gnosisEasyAuction, 0);
+        IERC20(auctionDetails.oTokenAddress).safeApprove(
+            gnosisEasyAuction,
+            IERC20(auctionDetails.oTokenAddress).balanceOf(address(this))
+        );
```
kenchangh, line 50: Since it's a new otoken, it's safe to assume the approval for the auction contract should be `0`. One way to save gas is to check if the approval != 0 before doing the double approve

```solidity
if (IERC20(auctionDetails.oTokenAddress).approval(address(this), gnosisEasyAuction) > 0) {
  IERC20(auctionDetails.oTokenAddress).safeApprove(gnosisEasyAuction, 0);
}
```

contracts/protocols/GnosisAuction.sol
```diff
@@ -1,4 +1,114 @@
...
+        optionPremium = uint96(
+            IOptionsPremiumPricer(optionsPremiumPricer)
+                .getPremium(
+                underlying,
+                newOToken.strikePrice(),
+                newOToken.expiryTimestamp(),
+                newOToken.isPut()
+            )
+                .mul(premiumDiscount)
+                .div(1000)
```
kenchangh, line 105: Do a uint96 overflow check here too

contracts/vaults/RibbonThetaVault.sol
chuddster: yep, changed

contracts/vaults/RibbonThetaVault.sol
chuddster: 👍

contracts/vaults/RibbonThetaVault.sol
chuddster: yes. 

It is not internal because there should be the optionality to redo the auction for the **current** options for whatever unforseen circumstances may arise. For example, there could be some issue where we don't think we have enough oTokens sold (we can adjust this in initiateAuction to not sell at all if we do not reach a goal), and we might want to redo it again for some reason. There are other wild cases that I might not be thinking of that might be worth being ready for at the expense of 250 -> 500 gas. Keeping it internal means it can only be called once for the current option.

it is not external because it is called internally, so it is public

contracts/vaults/RibbonThetaVault.sol
chuddster: Yeah, that is how I originally designed it. I decoupled to save gas. can do an offchain check to see if we even _need_ to burn remaining oTokens. but on second thought, perhaps convencience of one tx >> if statement gas

contracts/protocols/GnosisAuction.sol
chuddster: 👍

contracts/protocols/GnosisAuction.sol
chuddster: 👍

contracts/vaults/RibbonThetaVault.sol
kenchangh: Cool, this design of making the auction redo-able is interesting and potentially useful. My feedback on this,

* We should make the `startAuction` function permission-less (remove onlyManager) because it just kicks off the options process. (I don't see a reason why we need it to be permissioned)
* Can we get better guarantees around how often an auction will be kicked off? Let's say we start an auction of 100 otokens, we only sell off 80 otokens. We want to burn the remaining 20 otokens. But when we post the tx to burn the otokens, we get frontrun by bots that call `startAuction` again. I'm not sure what the solution here is, just putting out the possibility of a DoS attack where we are not able to ever burn the otokens.

contracts/vaults/RibbonThetaVault.sol
kenchangh: Yea let's stick with 2 different functions, 1 for closing the auction, another for burning otokens.

chuddster: closing the auction = IERC20(GNOSIS_EASY_AUCTION).settleAuction(id)

we can just call this function directly on gnosis contract, don't think we should have a proxy for it in theta vault code

contracts/vaults/RibbonThetaVault.sol
chuddster: 👍

PR #3, opened 2021-06-05
Vault deposit function

kenchangh:
PR has a basic version of the contract working with deposit. Changes include:

* Working `deposit` function. New deposits are queued per-address as a `DepositReceipt` object.
* Removed `withdraw` and `withdrawETH` function as we're using `withdrawLater` moving forward
* Remove fee payment from calculation and `completeScheduledWithdrawal` function
* Commented out the failing tests for merge
* Renamed lib/ to vendor/

**How deposits work**

1. User calls `deposit` function
2. If the user does not have a DepositReceipt, we create one with the current round and amount.
2. If the user already has a DepositReceipt for the current round, we top it up with the new deposit


contracts/vaults/RibbonThetaVault.sol
```diff
@@ -158,10 +155,11 @@ contract RibbonThetaVault is DSMath, GnosisAuction, OptionsVaultStorage {
         isPut = _isPut;
 
         // hardcode the initial withdrawal fee
-        instantWithdrawalFee = 0.005 ether;
+        instantWithdrawalFee = 0 ether;
```
kenchangh, line 158: Leaving this as 0 ether for now, will remove withdrawal fee logic completely in a follow-up PR

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -286,89 +268,66 @@ contract RibbonThetaVault is DSMath, OptionsVaultStorage {
...
+        IERC20(asset).safeTransferFrom(msg.sender, address(this), amount);
     }
 
     /**
      * @notice Mints the vault shares to the msg.sender
      * @param amount is the amount of `asset` deposited
      */
     function _deposit(uint256 amount) private {
-        uint256 totalWithDepositedAmount = totalBalance();
+        uint16 _round = round;
```
chuddster, line 294: where is `round` updated? should it be incremented in rollToNextOption?

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -286,89 +268,66 @@ contract RibbonThetaVault is DSMath, OptionsVaultStorage {
...
-     * @notice Burns vault shares and checks if eligible for withdrawal
-     * @param share is the number of vault shares to be burned
-     */
-    function _withdraw(uint256 share) private returns (uint256) {
-        (uint256 amountAfterFee, uint256 feeAmount) =
-            withdrawAmountWithShares(share);
+        // If we have a pending deposit in the current round, we add on to the pending deposit
+        if (_round == depositReceipt.round) {
+            // No deposits allowed until the next round
+            require(!depositReceipt.processed, "Processed");
```
chuddster, line 312: so receipts will **only** be processed after weeks shorts are closed?

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -286,89 +268,66 @@ contract RibbonThetaVault is DSMath, OptionsVaultStorage {
...
-        // solhint-disable-next-line
-        // https://github.com/AlphaFinanceLab/alphahomora/blob/340653c8ac1e9b4f23d5b81e61307bf7d02a26e8/contracts/5/Bank.sol#L104
-        uint256 share =
-            shareSupply == 0 ? amount : amount.mul(shareSupply).div(total);
-
-        require(shareSupply.add(share) >= minimumSupply, "Insufficient supply");
-
-        emit Deposit(msg.sender, amount, share);
-
-        _mint(msg.sender, share);
```
chuddster, line 333: so are [deposits](https://www.notion.so/ribbonfinance/Updated-Vault-Design-29df965a02b7455e9f6b18cb22cfbcf7#109b56f4ce394da6b3ffc9af6b0a04b7) not minting shares?

its either custodial OR non-custodial unlock + withdraw burn?

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -395,21 +354,18 @@ contract RibbonThetaVault is DSMath, OptionsVaultStorage {
...
         scheduledWithdrawals[msg.sender] = 0;
         queuedWithdrawShares = queuedWithdrawShares.sub(withdrawShares);
+        uint256 withdrawAmount = withdrawAmountWithShares(withdrawShares);
 
-        (uint256 amountAfterFee, uint256 feeAmount) =
-            withdrawAmountWithShares(withdrawShares);
-
-        emit Withdraw(msg.sender, amountAfterFee, withdrawShares, feeAmount);
-        emit ScheduledWithdrawCompleted(msg.sender, amountAfterFee);
+        emit Withdraw(msg.sender, withdrawAmount, withdrawShares);
```
chuddster, line 360: we need to add logic in withdraw later for checking whether transfer of shares to vault is needed. if they opt out of self-custody they wont have any shares.

test/RibbonThetaVault.ts
```diff
@@ -1134,1263 +1092,787 @@ function behavesLikeRibbonOptionsVault(params: {
...
-        const depositAmount = BigNumber.from("100000000000");
-
-        await depositIntoVault(params.collateralAsset, vault, depositAmount);
-
-        // simulate setting a bad otoken
-        await vault.connect(managerSigner).commitAndClose();
-
-        // users should have time to withdraw
-        await vault.withdraw(depositAmount.sub(await vault.minimumSupply()));
-      });
```
chuddster, line 2313: should we just do it.skip() and then change tests?

contracts/vaults/RibbonThetaVault.sol
kenchangh: Yep, it will be incremented in rollToNextOption, it's in the next PR

contracts/vaults/RibbonThetaVault.sol
kenchangh: They can only be processed in the new round. There are a few ways to get them processed:
- New deposit in the next round
- User calls `redeemDeposit` manually to redeem their shares

contracts/vaults/RibbonThetaVault.sol
kenchangh: Yep, that's right, deposits no longer mint shares

test/RibbonThetaVault.ts
kenchangh: Will do

contracts/vaults/RibbonThetaVault.sol
kenchangh: Noted, can I address this in a following PR?

test/RibbonThetaVault.ts
kenchangh: Actually, do you mind ignoring this for now, will be doing so in the next PR which includes the rollToNextOption tests

chuddster (approved):

PR #8, opened 2021-06-09
add strike selection changes

chuddster:


contracts/utils/StrikeSelection.sol
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /**
+     * @notice Gets the strike price satisfying the delta value
+     * given the expiry timestamp and whether option is call or put
+     * @param expiryTimestamp is the unix timestamp of expiration
+     * @param isPut is whether option is put or call
+     */
+
+    function getStrikePrice(uint256 expiryTimestamp, bool isPut)
```
chuddster, line 44: decimals here not entirely accurate - will adjust when I run tests

contracts/utils/StrikeSelection.sol
```diff
@@ -0,0 +1,101 @@
...
+
+        // For each asset prices with step of 'margin' (down if put, up if call)
+        //   if that assets getOptionDelta(currStrikePrice, t) == (isPut ? 1 - delta : delta) with certain margin of error
+        //        return strike price
+
+        bool pastDelta = false;
+        uint256 currStrike = assetPrice;
+        uint256 newDelta = isPut ? uint256(10).sub(delta) : delta;
+
+        while (!pastDelta) {
```
kenchangh, line 62: Love the approach on how to get strike price! A few comments:

- We should bound the number of iterations with a for loop instead of using a while loop. I haven't verified but it seems possible end up in an infinite loop
- Maybe fallback to whatever's the best strike price?

contracts/utils/StrikeSelection.sol
```diff
@@ -0,0 +1,101 @@
...
+                optionsPremiumPricer.getOptionDelta(currStrike, t);
+            if (
+                newDelta.sub(step.div(2).mul(newDelta)) <= currDelta &&
+                currDelta <= newDelta.add(step.div(2).mul(newDelta))
+            ) {
+                strikePrice = currStrike;
+            }
+            currStrike = isPut
+                ? currStrike.sub(currStrike.mul(step).div(1000))
+                : currStrike.add(currStrike.mul(step).div(1000));
```
kenchangh, line 73: One desirable thing we should aim for is to select strike prices that have similar increments (+100 for ETH, +1000 for BTC) as Deribit, this allows people to easily hedge on Deribit

image

The loop should iterate through the list with increments of +100: 3000, 3100, 3200, 3300

contracts/utils/StrikeSelection.sol
chuddster: So as it stands now, the delta will always be decreasing for calls and it will be increasing for puts. For calls we are looking for 0.1d and for puts 0.9d. When we iterate at each step we are either increasing or decreasing the delta. At some point we will reach the threshold of 0.1 and 0.9 and currDelta > newDelta will be true for puts or currDelta < newDelta will be true for calls.

We can add logic to just do % from spot price if while loop goes past x iterations? the only way this is possible is if chainlink asset oracle fails and latestAnswer() gives a wacky answer

contracts/utils/StrikeSelection.sol
chuddster: so do you want to just make step an absolute term that can be manually changed? ex: setStep -> 100 for eth / setStep -> 1000. then if eth appreciates we might do setStep -> 200 for eth based on what deribit does

kenchangh: Will leave that up to you, as long as we can minimize the iterations and gas use, I'm ok with either approach

contracts/utils/StrikeSelection.sol
kenchangh: Thanks for the clarification, makes sense that the iteration is bounded even with a while loop because it's just monotonically increasing/decreasing

contracts/utils/StrikeSelection.sol
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * @notice Gets the strike price satisfying the delta value
+     * given the expiry timestamp and whether option is call or put
+     * @param expiryTimestamp is the unix timestamp of expiration
+     * @param isPut is whether option is put or call
+     */
+
+    function getStrikePrice(uint256 expiryTimestamp, bool isPut)
+        external
+        returns (uint256 strikePrice)
```
kenchangh, line 46: Would be helpful to return `(strikePrice, delta)` so the consumer can emit the delta value out in an event for transparency

chuddster: 👍

contracts/utils/StrikeSelection.sol
chuddster: Yeah...think it makes sense to leave as is. Even if we do in absolute terms will still be imprecise (ex current price is 2374 and we have increments of 100 will be 2474, 2574 so wont match perfectly either way.) Can always modulate iterations by increasing step

kenchangh: If we iterate to 2474, we can always do a modulo so that it resolves correctly to 2400

`strike = strike - (strike % 100)`

test/StrikeSelection.ts
```diff
@@ -0,0 +1,123 @@
...
+      const isPut = false;
+      const targetDelta = await strikeSelection.delta();
+      const [strikePrice, delta] = await strikeSelection.getStrikePrice(
+        expiryTimestamp,
+        isPut
+      );
+      assert.equal(
+        strikePrice,
+        underlyingPrice.add(deltaAtUnderlying.sub(targetDelta).div(5).mul(100))
+      );
```
kenchangh, line 104: Add test for tie breaks

contracts/utils/StrikeSelection.sol
```diff
@@ -5,13 +5,128 @@ pragma experimental ABIEncoderV2;
...
+        // asset price
+        uint256 assetPrice = optionsPremiumPricer.getUnderlyingPrice();
+
+        // For each asset prices with step of 'margin' (down if put, up if call)
+        //   if asset's getOptionDelta(currStrikePrice, t) == (isPut ? 1 - delta:delta)
+        //   with certain margin of error
+        //        return strike price
+
+        bool pastDelta = false;
+        uint256 strike = assetPrice;
```
kenchangh, line 71: One easy fix to round out the strike prices to be like Deribit is to modulo the starting `strike`

```solidity
uint256 strike = assetPrice.sub(assetPrice % step);
```

So if we read 2742 as the `assetPrice`, we would set `strike = 2700` and iterate +-100 from there

contracts/utils/StrikeSelection.sol
```diff
@@ -5,13 +5,126 @@ pragma experimental ABIEncoderV2;
...
+                uint256 upperBoundDiff =
+                    isPut
+                        ? sub(currDelta, targetDelta)
+                        : sub(prevDelta, targetDelta);
+                uint256 lowerBoundDiff =
+                    isPut
+                        ? sub(targetDelta, prevDelta)
+                        : sub(targetDelta, currDelta);
+                // for tie breaks (ex: 0.05 <= 0.1 <= 0.15) round to higher strike price for calls and lower strike price for puts
+                return (
```
kenchangh, line 95: Before we return, we should do a `require(isPut ? strike < assetPrice : strike > assetPrice)` so we have a final sanity check

kenchangh (dismissed):
Changes LGTM, let's fix the tests before merging

kenchangh (approved):

PR #9, opened 2021-06-10
Working commitAndClose and rollToNextOption

kenchangh:


contracts/vaults/RibbonThetaVault.sol
```diff
@@ -318,18 +325,66 @@ contract RibbonThetaVault is DSMath, OptionsVaultStorage {
...
+            depositReceipts[msg.sender];
+        _redeemDeposit(round, depositReceipt);
+    }
+
+    function _redeemDeposit(
+        uint16 currentRound,
+        VaultDeposit.DepositReceipt memory depositReceipt
+    ) private {
+        require(!depositReceipt.processed, "Processed");
+        require(depositReceipt.round > currentRound, "Round not closed");
```
chuddster, line 358: should it be `depositReceipt.round < currentRound` based on [this](https://github.com/ribbon-finance/ribbon-v2/pull/3#discussion_r649892802)

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -318,18 +325,66 @@ contract RibbonThetaVault is DSMath, OptionsVaultStorage {
...
+            depositReceipts[msg.sender];
+
+        require(!depositReceipt.processed, "Processed");
+        require(depositReceipt.round == round, "Invalid round");
+
+        // Subtraction underflow checks already ensure it is smaller than uint128
+        depositReceipt.amount = uint128(
+            uint256(depositReceipt.amount).sub(amount)
+        );
+
```
chuddster, line 386: would it be cheaper to do depositReceipts[msg.sender].processed, etc everywhere instead of creating storage variable pointer depositReceipt?

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -412,11 +467,11 @@ contract RibbonThetaVault is DSMath, OptionsVaultStorage {
...
         _closeShort(oldOption);
-    }
 
-    function closeShort() external nonReentrant {
-        address oldOption = currentOption;
-        _closeShort(oldOption);
+        // After closing the short, if the options expire in-the-money
+        // vault pricePerShare would go down because vault's asset balance decreased.
+        // This ensures that the newly-minted shares do not take on the loss.
+        _mintPendingShares();
```
chuddster, line 474: when you say "newly minted shares" do you mean people who deposited in the current round? So you rebalance the price per share by inflating the price per share (by minting) so that the current round's depositors withdrawal amount will not take into account losses?

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -491,8 +553,6 @@ contract RibbonThetaVault is DSMath, OptionsVaultStorage {
             shortAmount
         );
 
-        startAuction();
-
```
chuddster, line 495: you think we should start the auction in a separate tx?

contracts/vaults/RibbonThetaVault.sol
kenchangh: Yep, was a typo

contracts/vaults/RibbonThetaVault.sol
kenchangh: Do you mean `depositReceipts[msg.sender].amount`? There isn't a difference because it's just a single `SSTORE` opcode

contracts/vaults/RibbonThetaVault.sol
kenchangh: Not sure what you mean by "rebalancing the price per share". The process of minting and burning doesn't actually affect the price per share.

Let me give you an example, when `balance = 100`, `shareSupply = 100`, pricePerShare will be `1`.

```js
pricePerShare = balance / shareSupply
```

 When you mint new shares, you don't actually update the pricePerShare value at all.

```js
// We mint
balance += 10
sharesMinted = pricePerShare * balance
shareSupply += sharesMinted
// pricePerShare remains the same!
```

kenchangh: In the case of minting after closing the short, we can think of it as "let's use the pricePerShare right *after* the vault takes the loss".

Starting scenario:
```js
balance = 100
shareSupply = 100
pricePerShare = balance / shareSupply // pricePerShare = 1
```

After closing the short:
```js
closeShort() // causes balance to be -10, so 90
pricePerShare = balance / shareSupply // 90/100 = 0.9

mintShares() // mint shares using the new pricePerShare
```

contracts/vaults/RibbonThetaVault.sol
kenchangh: Removed this by accident, will add back

contracts/vaults/RibbonThetaVault.sol
chuddster: yep! makes sense

chuddster (approved):

PR #10, opened 2021-06-10
Working redeem deposit

kenchangh:
### Recap

1. When users deposit, we create a `DepositReceipt` object for them. This stores a reference to how much they deposited.
2. When the `rollToNextOption` function is called, all pending deposit amounts (the sum of all deposits) are processed (i.e. minted vault shares).
3. Users do not actually hold ownership of the shares. They are held on the vault address itself.

### Explicit redemptions

Users can call the `redeemDeposit` function to release the owed shares to their account. It refers to the pricePerShare stored in the `roundPricePerShare` to calculate how many shares to release to the user.

The use case for an explicit redemption are for protocol integrations that want to have custody over the vault shares. (Composability)

### Implicit redemptions (via new deposits)

When the user makes a deposit in a new round, we implicitly return the vault shares from the prior rounds to their account.

There are a few conditions for an implicit redemption:
1. User has a DepositReceipt from prior round
2. User has a DepositReceipt that is unprocessed


contracts/vaults/RibbonThetaVault.sol
```diff
@@ -312,7 +313,7 @@ contract RibbonThetaVault is DSMath, OptionsVaultStorage {
 
         emit Deposit(msg.sender, amount, currentRound);
 
-        VaultDeposit.DepositReceipt storage depositReceipt =
+        VaultDeposit.DepositReceipt memory depositReceipt =
```
chuddster, line 316: doesn't memory create a new copy? 

https://www.geeksforgeeks.org/storage-vs-memory-in-solidity/

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -536,26 +533,57 @@ contract RibbonThetaVault is DSMath, OptionsVaultStorage {
         address newOption = nextOption;
         require(newOption > PLACEHOLDER_ADDR, "!nextOption");
 
+        uint256 pendingAmount = totalPending();
+        uint256 currentSupply = totalSupply();
+        uint256 currentBalance = assetBalance();
+        uint256 roundStartBalance = currentBalance.sub(pendingAmount);
+        uint16 currentRound = round;
```
chuddster, line 540: perhaps easier to remove `uint16 currentRound = round;` and do `round += 1;` after `roundPricePerShare[currentRound] = currentPricePerShare;`

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -536,26 +533,57 @@ contract RibbonThetaVault is DSMath, OptionsVaultStorage {
         address newOption = nextOption;
         require(newOption > PLACEHOLDER_ADDR, "!nextOption");
```
chuddster, line 534: so newOption > 1 is same as newOption != address(0)?

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -536,26 +533,57 @@ contract RibbonThetaVault is DSMath, OptionsVaultStorage {
...
+            currentSupply > 0
+                ? singleShare.mul(roundStartBalance).div(currentSupply)
+                : singleShare;
+
+        // After closing the short, if the options expire in-the-money
+        // vault pricePerShare would go down because vault's asset balance decreased.
+        // This ensures that the newly-minted shares do not take on the loss.
+        uint256 mintShares =
+            currentSupply > 0
+                ? pendingAmount.mul(currentSupply).div(roundStartBalance)
```
chuddster, line 554: can do `uint256 mintShares = pendingAmount.div(currentPricePerShare).mul(singleShare)`

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -536,26 +533,57 @@ contract RibbonThetaVault is DSMath, OptionsVaultStorage {
...
+
+        // Vault holds temporary custody of the newly minted vault shares
+        _mint(address(this), mintShares);
+
+        uint256 newSupply = currentSupply.add(mintShares);
+
+        // We assume that users cannot queued withdrawals in the first round
+        uint256 queuedWithdrawAmount =
+            newSupply > 0
+                ? queuedWithdrawShares.mul(currentBalance).div(newSupply)
```
chuddster, line 565: wanted to resurface the issue with queuedWithdrawShares where when they completeWithdrawals they will be getting it at the pps of that round, not the round in which they initiated withdraw.

[our convo](https://discord.com/channels/@me/814424663551377439/852989414787121202)

contracts/vaults/RibbonThetaVault.sol
kenchangh: Yep!

contracts/vaults/RibbonThetaVault.sol
kenchangh: Will add a TODO comment on this

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -536,26 +533,60 @@ contract RibbonThetaVault is DSMath, OptionsVaultStorage {
...
 
-        uint256 currentBalance = assetBalance();
-        (uint256 queuedWithdrawAmount, , ) =
-            _withdrawAmountWithShares(queuedWithdrawShares, currentBalance);
-        uint256 shortAmount = currentBalance.sub(queuedWithdrawAmount);
-        lockedAmount = shortAmount;
+        // Finalize the pricePerShare at the end of the round
+        uint16 currentRound = round;
+        roundPricePerShare[currentRound] = currentPricePerShare;
+        round = currentRound + 1;
```
kenchangh, line 578: The `uint16 currentRound = round` is needed because it helps us save ~100 gas from avoiding a read

chuddster: understood

contracts/vaults/RibbonThetaVault.sol
chuddster: so the ternary operation is more gas effective than no ternary operation + potentially pendingAmount.mul(singleShare).div(singleShare)?

kenchangh: I overlooked the ternary, I [removed it](https://github.com/ribbon-finance/ribbon-v2/pull/10/commits/61a5f519ffc12b01f897d37fcdfad355b68f8eb3)

chuddster (approved):

PR #11, opened 2021-06-12
Add snapshotted premium pricing before auction, override strike selection

chuddster:


contracts/vaults/RibbonThetaVault.sol
```diff
@@ -390,8 +390,15 @@ contract RibbonThetaVault is DSMath, OptionsVaultStorage {
             expiry = getNextFriday(IOtoken(oldOption).expiryTimestamp());
         }
 
+        IStrikeSelection strikeSelection = IStrikeSelection(strikeSelection);
+
         (uint256 strikePrice, uint256 delta) =
-            IStrikeSelection(strikeSelection).getStrikePrice(expiry, isPut);
+            lastStrikeOverride == round
+                ? (overridenStrikePrice, strikeSelection.delta())
```
chuddster, line 397: Should not have round = 0 because then for first v2 round will use overridden. we should start with round = 1

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -390,8 +390,15 @@ contract RibbonThetaVault is DSMath, OptionsVaultStorage {
...
+                : IStrikeSelection(strikeSelection).getStrikePrice(
+                    expiry,
+                    isPut
+                );
```
chuddster, line 402: Should not have round = 0 because then for first v2 round will use overridden. we should start with round = 1

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -410,6 +417,15 @@ contract RibbonThetaVault is DSMath, OptionsVaultStorage {
 
         emit NewOptionStrikeSelected(strikePrice, delta);
 
+        currentOtokenPremium = GnosisAuction.getOTokenPremium(
+            otokenAddress,
+            GNOSIS_EASY_AUCTION,
+            optionsPremiumPricer,
+            premiumDiscount
```
chuddster, line 424: snapshots premium in commitAndClose and stores

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -531,6 +545,20 @@ contract RibbonThetaVault is DSMath, OptionsVaultStorage {
...
+    function setStrikePrice(uint256 strikePrice)
+        external
+        onlyManager
+        nonReentrant
+    {
+        require(strikePrice > 0, "!strikePrice");
+        overridenStrikePrice = strikePrice;
+        lastStrikeOverride = round;
+    }
+
```
chuddster, line 561: 1. updates overridenStrikePrice with new strike price
2. updates latest round which we overrode strike price. in commitAndClose if that latest round is this round then we use manual instead of strike selection contract methodology

this needs to be called before commitAndClose. so off-chain check of whether strike selection contract is acting up should be beforehand

contracts/protocols/GnosisAuction.sol
```diff
@@ -22,46 +22,45 @@ library GnosisAuction {
...
         }
 
         IERC20(auctionDetails.oTokenAddress).safeApprove(
-            gnosisEasyAuction,
+            auctionDetails.gnosisEasyAuction,
             IERC20(auctionDetails.oTokenAddress).balanceOf(address(this))
         );
 
+        require(
+            auctionDetails.oTokenPremium.mul(oTokenSellAmount) <=
```
kenchangh, line 57: We can store the result of the multiplication in a variable so we're not doing it again (it has the SafeMath overhead)

contracts/protocols/GnosisAuction.sol
```diff
@@ -73,7 +72,7 @@ library GnosisAuction {
                 // we are selling all of the otokens minus a fee taken by gnosis
                 uint96(oTokenSellAmount),
                 // the minimum we are willing to sell all the oTokens for. A discount is applied on black-scholes price
-                uint96(optionPremium.mul(oTokenSellAmount)),
+                uint96(auctionDetails.oTokenPremium.mul(oTokenSellAmount)),
```
kenchangh, line 75: This is what I was referring to https://github.com/ribbon-finance/ribbon-v2/pull/11/files#r650442961

contracts/storage/OptionsVaultStorage.sol
```diff
@@ -86,6 +86,15 @@ contract OptionsVaultStorageV3 {
     /// @notice Logic contract used to select strike prices
     address public strikeSelection;
 
+    /// @notice Last `round` when strike overriden
+    uint256 public lastStrikeOverride;
+
+    /// @notice Current overriden strike price
+    uint256 public overridenStrikePrice;
```
kenchangh, line 93: One way to optimize for storage read/writes here is to use a struct to pack these variables together. This can be done because `lastStrikeOverride` is a `uint16` number. See how we do it in [DepositReceipt](https://github.com/ribbon-finance/ribbon-v2/blob/master/contracts/libraries/VaultDeposit.sol#L5)

chuddster: so I am guessing in deposit receipt this works because it crams round (uint16) + amount (uint128) into a uint256 number which does not take as much space as either of them separate. 

so in my case this would only make sense if overridenStrikePrice will fit in uint128 - we can confirm this right

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -390,8 +390,18 @@ contract RibbonThetaVault is DSMath, OptionsVaultStorage {
...
         }
 
+        IStrikeSelection strikeSelection = IStrikeSelection(strikeSelection);
+
         (uint256 strikePrice, uint256 delta) =
-            IStrikeSelection(strikeSelection).getStrikePrice(expiry, isPut);
+            strikeOverride.lastStrikeOverride == round
+                ? (
+                    strikeOverride.overriddenStrikePrice,
+                    strikeSelection.delta()
```
chuddster, line 399: in the strike override we assume the strike price will be 0.1 delta, or whatever the delta is in strike selection contract. can we make this assumption or should we add an additional param to setStrikePrice that sets some `overrideDelta` variable.

kenchangh: Seems reasonable to just return the `strikeSelection.delta()`. Don't think the delta matters in that scenario

test/RibbonThetaVault.ts
```diff
@@ -234,26 +234,27 @@ function behavesLikeRibbonOptionsVault(params: {
...
+
   // Variables
   let defaultOtokenAddress: string;
 
-  describe(`${params.name}`, () => {
-    let initSnapshotId: string;
-    let firstOption: Option;
-    let secondOption: Option;
+  const rollToNextOption = async () => {
+    await strikeSelection.setStrikePrice(
```
chuddster, line 244: need to fix tests later

contracts/storage/OptionsVaultStorage.sol
kenchangh: Yep that's right, you can  read more about [struct packing here](https://fravoll.github.io/solidity-patterns/tight_variable_packing.html)

For every 256 bits, the EVM allocates a storage slot. As long as we stayed <= 256 we can save on using an extra slot

kenchangh (approved):

PR #16, opened 2021-06-17
Updated redemption model

kenchangh:
This PR changes the model of how we handle redemptions. Previously we did an implicit redemption that transfers the users shares when they do a new deposit.

Instead of transferring tokens implicitly (which would be an accounting mess), we would track their unredeemed share balance on `DepositReceipt.unredeemedShares`.

We are changing it to:
1. User deposits into vault
2. Vault is rolled, mints shares and holds it on its balance.
3. When the user creates a new deposit, the `unredeemedShares` is updated to include the previous round's `amount`.

Changes include:
* Remove implicit redemption in `deposit`
* `redeem` and `maxRedeem` functions
* `shares` getter that gives the user's share balance, inclusive of what is owed to them (used by FE)
* Removed unused dead code

### DIFF
--- a/contracts/vaults/RibbonThetaVault.sol
+++ b/contracts/vaults/RibbonThetaVault.sol
@@ -36,9 +36,6 @@ contract RibbonThetaVault is DSMath, OptionsVaultStorage {
     address public immutable WETH;
     address public immutable USDC;
 
-    // 90% locked in options protocol, 10% of the pool reserved for withdrawals
-    uint256 public constant lockedRatio = 0.9 ether;
-
     uint256 public constant delay = 1 hours;
 
     uint256 public constant period = 7 days;
@@ -720,6 +717,19 @@ contract RibbonThetaVault is DSMath, OptionsVaultStorage {
     function pricePerShare() external view returns (uint256) {
         uint256 balance = totalBalance().sub(totalPending());
         return (10**uint256(_decimals)).mul(balance).div(totalSupply());
+    }
+
+    /**
+     * @notice Returns the expiry of the current option the vault is shorting
+     */
+    function currentOptionExpiry() external view returns (uint256) {
+        address _currentOption = currentOption;
+        if (_currentOption == address(0)) {
+            return 0;
+        }
+
+        IOtoken oToken = IOtoken(_currentOption);
+        return oToken.expiryTimestamp();
     }
 
     /**
--- a/contracts/vaults/RibbonThetaVault.sol
+++ b/contracts/vaults/RibbonThetaVault.sol
@@ -385,17 +385,15 @@ contract RibbonThetaVault is DSMath, OptionsVaultStorage {
         shares = isMax ? unredeemedShares : shares;
         require(shares <= unredeemedShares, "Exceeds available");
 
-        if (depositReceipt.processed) {
+        // When the depositReceipt is processed, we need subtract the shares being redeemed
+        // from the `unredeemedShares`.
+        // Alternatively, if we try to redeem more than the current round's shares
+        // We need to subtract the additional shares from `unredeemedShares`.
+        if (depositReceipt.processed || shares > roundShares) {
             depositReceipts[msg.sender].unredeemedShares = uint128(
                 uint256(depositReceipt.unredeemedShares).sub(shares)
             );
-        } else {
-            if (shares > roundShares) {
-                depositReceipts[msg.sender].unredeemedShares = uint128(
-                    uint256(depositReceipt.unredeemedShares).sub(shares)
-                );
-            }
-
+        } else if (!depositReceipt.processed) {
             depositReceipts[msg.sender].processed = true;
         }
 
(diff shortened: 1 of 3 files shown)


contracts/vaults/RibbonThetaVault.sol
```diff
@@ -321,76 +318,128 @@ contract RibbonThetaVault is DSMath, OptionsVaultStorage {
...
+    function redeem(uint256 shares) external nonReentrant {
+        require(shares > 0, "!shares");
+        _redeem(shares, false);
     }
 
-    function redeemDeposit() external nonReentrant {
+    /**
+     * @notice Redeems the entire unredeemedShares balance that is owed to the account
+     */
+    function maxRedeem() external nonReentrant {
```
chuddster, line 370: maybe just have a redeem() and in the frontend we have max button + total unredeemed calculation off-chain?

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -321,76 +318,128 @@ contract RibbonThetaVault is DSMath, OptionsVaultStorage {
...
+        require(shares <= unredeemedShares, "Exceeds available");
 
-        uint256 shares =
-            uint256(depositReceipt.amount).mul(10**uint256(_decimals)).div(pps);
+        depositReceipts[msg.sender] = VaultDeposit.DepositReceipt({
+            processed: true,
+            round: depositReceipt.round,
+            amount: depositReceipt.amount,
+            unredeemedShares: uint128(uint256(unredeemedShares).sub(shares))
+        });
```
chuddster, line 400: wouldn't changing processed -> true and unredeemedShares -> unredeemedShares - x

be more gas effective than reassigning entire struct?

kenchangh: Will test this out, from first glance it should be same because the struct is packed into a single storage slot

contracts/vaults/RibbonThetaVault.sol
kenchangh: I think the redemption functionality is mostly useful for other contracts. For users that use the FE, they don't have to call the `redeem` function in the first place.

If we're writing this function to cater towards contracts, they would appreciate a `maxRedeem` function because it's more gas efficient

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -321,76 +318,128 @@ contract RibbonThetaVault is DSMath, OptionsVaultStorage {
...
+            require(pps > PLACEHOLDER_UINT, "Invalid pps");
+
+            uint256 sharesFromRound =
+                uint256(depositReceipt.amount).mul(10**uint256(_decimals)).div(
+                    pps
+                );
+            require(sharesFromRound < type(uint104).max, "Overflow");
+
+            uint256 unredeemedShares256 =
+                uint256(depositReceipt.unredeemedShares).add(sharesFromRound);
```
chuddster, line 430: is there a possible double counting here? sharesFromRound takes in `depositReceipt.amount` but the deposit amount respective shares could already be reflected in the existing `depositReceipt.unredeemedShares` so we would be adding depositReceipt.unredeemedShares + sharesFromRound and therefore doublecounting?

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -399,56 +448,17 @@ contract RibbonThetaVault is DSMath, OptionsVaultStorage {
...
         emit InstantWithdraw(msg.sender, amount, currentRound);
 
         transferAsset(msg.sender, amount);
     }
 
-    /**
-     * @notice Lock's users shares for future withdraw and ensures that the new short excludes the scheduled amount.
-     * @param shares is the number of shares to be withdrawn in the future.
-     */
-    function withdrawLater(uint256 shares) external nonReentrant {
```
chuddster, line 417: why exactly we removing this stuff? are you just going to put it back in with the subsequent withdraw later pr?

contracts/vaults/RibbonThetaVault.sol
chuddster: perhaps from a storage perspective it is the same. but I meant in terms of opcodes with reassigning struct as opposed to uints

contracts/vaults/RibbonThetaVault.sol
kenchangh: I was running into bytecode size problems so removing them was a hack haha. Plan to add them back

contracts/vaults/RibbonThetaVault.sol
kenchangh: I tested it and your suggestion was better on gas

contracts/vaults/RibbonThetaVault.sol
kenchangh: There wasn't an issue with double counting because we check if `depositReceipt.processed` is true. But I made the change to zero out the `depositReceipt.amount` whenever [we do a redemption](https://github.com/ribbon-finance/ribbon-v2/pull/16/commits/7d701ba2c8a4fa9aec93cf84f10f5b410e16a9dd#diff-f882ad74759f8aa2abee31cab887af57224330691be48e0a1e7b1a66f2075cbaR396) so we have extra guarantees

chuddster (approved):

PR #18, opened 2021-06-17
Optimize bytecode size, remove manager

kenchangh:
This change decreases the bytecode size from 22.8KB to 22.2KB

- Decrease reason string length
- Refactor a few functions
- Remove the manager role entirely, just having the owner is enough

### DIFF
--- a/contracts/storage/OptionsVaultStorage.sol
+++ b/contracts/storage/OptionsVaultStorage.sol
@@ -25,7 +25,7 @@ contract OptionsVaultStorageV1 is
     address private _oldAsset;
 
     // Privileged role that is able to select the option terms (strike price, expiry) to short
-    address public manager;
+    address private _manager;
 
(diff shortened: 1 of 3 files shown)


contracts/storage/OptionsVaultStorage.sol
```diff
@@ -25,7 +25,7 @@ contract OptionsVaultStorageV1 is
     address private _oldAsset;
 
     // Privileged role that is able to select the option terms (strike price, expiry) to short
-    address public manager;
+    address private _manager;
```
chuddster, line 28: do we need manager here?

kenchangh: Trying to keep the storage layout compatible with v1 atm, so we have to retain it

chuddster (approved):

PR #19, opened 2021-06-17
initiateWithdraw function

kenchangh:
* initiateWithdraw function that creates a new `Withdrawal` object per address
* `Withdrawal` data structure that takes note of the `round` the withdrawal was initiated in (used to reference `roundPricePerShare` later on)

### DIFF
--- a/contracts/storage/OptionsVaultStorage.sol
+++ b/contracts/storage/OptionsVaultStorage.sol
@@ -10,7 +10,7 @@ import {
 import {
     ERC20Upgradeable
 } from "@openzeppelin/contracts-upgradeable/token/ERC20/ERC20Upgradeable.sol";
-import {VaultDeposit} from "../libraries/VaultDeposit.sol";
+import {Vault} from "../libraries/Vault.sol";
 import {StrikeOverride} from "../libraries/StrikeOverride.sol";
 
 contract OptionsVaultStorageV1 is
@@ -18,15 +18,6 @@ contract OptionsVaultStorageV1 is
     OwnableUpgradeable,
     ERC20Upgradeable
 {
-    // DEPRECATED: This variable was originally used to store the asset address we are using as collateral
-    // But due to gas optimization and upgradeability security concerns,
-    // we removed it in favor of using immutable variables
-    // This variable is left here to hold the storage slot for upgrades
-    address private _oldAsset;
-
-    // Privileged role that is able to select the option terms (strike price, expiry) to short
-    address private _manager;
-
     // Option that the vault is shorting in the next cycle
     address public nextOption;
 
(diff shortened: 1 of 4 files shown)


contracts/storage/OptionsVaultStorage.sol
```diff
@@ -101,22 +88,21 @@ contract OptionsVaultStorageV3 {
...
 }
 
 // We are following Compound's method of upgrading new contract implementations
 // When we need to add new storage variables, we create a new version of OptionsVaultStorage
 // e.g. OptionsVaultStorageV<versionNumber>, so finally it would look like
 // contract OptionsVaultStorage is OptionsVaultStorageV1, OptionsVaultStorageV2
-contract OptionsVaultStorage is
-    OptionsVaultStorageV1,
-    OptionsVaultStorageV2,
-    OptionsVaultStorageV3
```
kenchangh, line 119: Decided to introduce incompatibility with Ribbon v1, we're going to do a migration of funds from v1 contract to v2 contract

test/RibbonThetaVault.ts
```diff
@@ -1857,225 +1857,128 @@ function behavesLikeRibbonOptionsVault(params: {
...
-
-    //     // Queued withdrawals + 10% of available assets set aside
-    //     assert.equal(
-    //       vaultBalanceBeforeWithdraw.toString(),
-    //       BigNumber.from("110000000000").toString()
-    //     );
-    //   });
-    // });
-
-    // describe("completeScheduledWithdrawal", () => {
```
kenchangh, line 1954: Removing and starting afresh

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -434,6 +435,40 @@ contract RibbonThetaVault is OptionsVaultStorage {
...
+     * @param shares is the number of shares to withdraw
+     */
+    function initiateWithdraw(uint128 shares) external nonReentrant {
+        require(shares > 0, "!shares");
+
+        // This caches the `round` variable used in shareBalances
+        uint16 currentRound = round;
+        Vault.Withdrawal memory withdrawal = withdrawals[msg.sender];
+
+        require(!withdrawal.initiated, "Existing withdraw");
```
chuddster, line 449: so we are constraining it to only one withdrawal per round?

ex:

initiateWithdraw(half of my shares)
10 min later
initiateWithdraw(quarter of my shares)

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -434,6 +435,40 @@ contract RibbonThetaVault is OptionsVaultStorage {
...
+
+        emit InitiateWithdraw(msg.sender, shares, currentRound);
+
+        withdrawals[msg.sender].initiated = true;
+        withdrawals[msg.sender].round = currentRound;
+        withdrawals[msg.sender].shares = shares;
+
+        queuedWithdrawShares = queuedWithdrawShares.add(shares);
+
+        if (shares > heldByVault) {
```
chuddster, line 466: shouldn't it be shares > heldByAccount? if the shares is greater than held by user then we will need to transfer shares from vault to user.

contracts/vaults/RibbonThetaVault.sol
kenchangh: Yep, we're limiting only 1 withdrawal per round. This is really for simplicity

chuddster (approved):

PR #24, opened 2021-06-22
fix decimal compatibility

chuddster:
1. shifts decimals of premium in GnosisAuction to be consistent with the bidding token (eth for eth call vaults, usdc for eth put vaults, etc)
3. make step, strike price, asset price in strikeSelection all scaled by 10 ** 8 for consistency with options premium pricer and oToken strikeprice decimals

### DIFF
--- a/contracts/libraries/GnosisAuction.sol
+++ b/contracts/libraries/GnosisAuction.sol
@@ -58,7 +58,7 @@ library GnosisAuction {
         // shift decimals to correspond to decimals of USDC for puts
         // and underlying for calls
         uint256 minBidAmount =
-            dswmul(oTokenSellAmount.mul(10**12), auctionDetails.oTokenPremium)
+            dswmul(oTokenSellAmount.mul(10**10), auctionDetails.oTokenPremium)
                 .div(10**(uint256(18).sub(auctionDetails.assetDecimals)));
 
         require(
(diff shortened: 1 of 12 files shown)


contracts/libraries/GnosisAuction.sol
```diff
@@ -53,8 +54,12 @@ library GnosisAuction {
             IERC20(auctionDetails.oTokenAddress).balanceOf(address(this))
         );
 
+        // minBidAmount is total oTokens to sell * premium per oToken
+        // shift decimals to correspond to decimals of USDC for puts
+        // and underlying for calls
         uint256 minBidAmount =
-            auctionDetails.oTokenPremium.mul(oTokenSellAmount);
+            dswmul(oTokenSellAmount.mul(10**12), auctionDetails.oTokenPremium)
+                .div(10**(uint256(18).sub(auctionDetails.assetDecimals)));
```
kenchangh, line 62: I'm trying to understand the math here. Why do we multiply 10**12 for the otoken amounts? Shouldn't it be 10**10 since otoken is 8 decimal places

contracts/utils/StrikeSelection.sol
```diff
@@ -37,7 +40,19 @@ contract StrikeSelection is DSMath, Ownable {
...
         delta = _delta;
         // ex: step = 1000
-        step = _step;
+        step = _step.mul(
+            10 **
+                IPriceOracle(
+                    IOptionsPremiumPricer(_optionsPremiumPricer).priceOracle()
+                )
+                    .decimals()
+        );
```
kenchangh, line 49: `_step.mul(assetOracleMultiplier)`?

kenchangh (approved):

PR #25, opened 2021-06-24
change access manager to address(0), fix placeSellOrder, make tests m…

chuddster:
…ore thorough

### DIFF
--- a/test/RibbonThetaVault.ts
+++ b/test/RibbonThetaVault.ts
@@ -103,11 +103,11 @@ describe("RibbonThetaVault", () => {
     secondOptionStrike: 64000,
     chainlinkPricer: CHAINLINK_WBTC_PRICER,
     tokenDecimals: 6,
-    depositAmount: BigNumber.from("100000000"),
-    premium: BigNumber.from("10000000"),
+    depositAmount: BigNumber.from("100000000000"),
+    premium: BigNumber.from("10000000000"),
     premiumDiscount: BigNumber.from("997"),
     minimumSupply: BigNumber.from("10").pow("3").toString(),
-    expectedMintAmount: BigNumber.from("158730"),
+    expectedMintAmount: BigNumber.from("158730158"),
     isPut: true,
     gasLimits: {
       depositWorstCase: 115000,
@@ -1116,7 +1116,7 @@ function behavesLikeRibbonOptionsVault(params: {
       });
     });
 
-    describe.skip("#burnRemainingOTokens", () => {
+    describe("#burnRemainingOTokens", () => {
       time.revertToSnapshotAfterEach(async function () {
         await depositIntoVault(params.collateralAsset, vault, depositAmount);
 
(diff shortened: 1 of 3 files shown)


test/RibbonThetaVault.ts
```diff
@@ -103,11 +103,11 @@ describe("RibbonThetaVault", () => {
     secondOptionStrike: 64000,
     chainlinkPricer: CHAINLINK_WBTC_PRICER,
     tokenDecimals: 6,
-    depositAmount: BigNumber.from("100000000"),
-    premium: BigNumber.from("10000000"),
+    depositAmount: BigNumber.from("100000000000"),
+    premium: BigNumber.from("10000000000"),
     premiumDiscount: BigNumber.from("997"),
     minimumSupply: BigNumber.from("10").pow("3").toString(),
```
chuddster, line 109: changed this to be consistent with eth theta vault put deposit amount for easier comparison

test/RibbonThetaVault.ts
```diff
@@ -1172,42 +1195,75 @@ function behavesLikeRibbonOptionsVault(params: {
...
-          parseInt(vaultOTokenBalanceAfter),
-          parseInt(vaultOTokenBalanceBefore)
+          parseInt((await defaultOtoken.balanceOf(vault.address)).toString()),
+          parseInt(
+            params.expectedMintAmount
+              .div(bidMultiplier)
+              .mul(params.premiumDiscount.sub(1))
+              .div(1000)
+              .toString()
+          )
```
chuddster, line 1232: If I do a placeSellOrder for say 5 tokens, and I am offering 1.5x than the minimum price per token, I actually get more than 5. In this case since we have a floor of premium * 997 / 1000 and my bidder actually pays premium, he gets ~3% more tokens so it will not exactly be 5 tokens. hence we do tokenBalanceAfterSettle > (mintAmount / 2 ) * 996 / 1000

test/RibbonThetaVault.ts
```diff
@@ -1172,42 +1195,75 @@ function behavesLikeRibbonOptionsVault(params: {
...
+        );
+
+        assert.isAbove(
+          parseInt((await assetContract.balanceOf(vault.address)).toString()),
+          parseInt(
+            (
+              (assetBalanceBeforeSettle.add(BigNumber.from(auctionDetails[2])) *
+                99) /
+              100
+            ).toString()
```
chuddster, line 1242: we are doing assetBalance > assetBalanceBefore * premiumGiven * 99 / 100 because we don't actually get back the premium we get a little less because of how the auction works. nothing to do with them taking fees. just that placeSellOrder(paying 1eth) does not actually pay 1 eth it ends up doing some funky stuff to calculate a smaller clearing price. If this is unconvincing, can dig a bit deeper :)

test/RibbonThetaVault.ts
```diff
@@ -1172,42 +1195,75 @@ function behavesLikeRibbonOptionsVault(params: {
...
+        );
 
         assert.isAbove(
-          parseInt(collateralBalanceAfter),
-          parseInt(collateralBalanceBefore)
+          parseInt(assetBalanceAfterBurn.toString()),
+          parseInt(
+            assetBalanceAfterSettle
+              .add(
+                lockedAmountBeforeBurn
```
chuddster, line 1260: similar to how [here](https://github.com/ribbon-finance/ribbon-v2/pull/25/files#r658355771) the person gets more tokens, this means we end up with _less_ tokens to burn, so a smaller balance afterwards (~3 less)

test/RibbonThetaVault.ts
kenchangh: This seems concerning - if I put in 100 ETH, do I get back 99 ETH in return? or 100 ETH - 1 wei.

If it's 100 ETH - 1 wei it's acceptable

kenchangh (approved):

test/RibbonThetaVault.ts
chuddster: it is more the latter.

PR #26, opened 2021-06-25
add E2E for strike selection

chuddster:


test/StrikeSelectionE2E.ts
```diff
@@ -0,0 +1,249 @@
...
+      );
+
+      assert.equal(
+        strikePrice.toString(),
+        underlyingPrice
+          .sub(
+            deltaAtUnderlying
+              .sub(targetDelta)
+              .div(500)
+              .sub(3)
```
chuddster, line 185: shifting by 300 here. seems counterintuitive, but say spot is 1800. 0.1 delta for call will be 2300 (diff of 500) and 0.1 delta for put will be 1500 (diff of 300). it makes sense that they would either be 500 from spot or 300 from spot. but they are not
Screen Shot 2021-06-25 at 1 57 53 PM



>

contracts/tests/TestVolOracle.sol
```diff
@@ -0,0 +1,72 @@
...
+import {SafeMath} from "@openzeppelin/contracts/math/SafeMath.sol";
+import {
+    PRBMathSD59x18
+} from "@ribbon-finance/rvol/contracts/libraries/PRBMathSD59x18.sol";
+import {Math} from "@ribbon-finance/rvol/contracts/libraries/Math.sol";
+import {DSMath} from "@ribbon-finance/rvol/contracts/libraries/DSMath.sol";
+import {Welford} from "@ribbon-finance/rvol/contracts/libraries/Welford.sol";
+import {VolOracle} from "@ribbon-finance/rvol/contracts/core/VolOracle.sol";
+
+contract TestVolOracle is DSMath, VolOracle {
```
kenchangh, line 13: We can just export this from rvol so we can use it our tests

test/RibbonThetaVault.ts
```diff
@@ -2615,6 +2657,46 @@ function behavesLikeRibbonOptionsVault(params: {
...
+      BigNumber.from("2250000000"),
+      BigNumber.from("2650000000"),
+    ];
+
+    for (let i = 0; i < values.length; i++) {
+      await volOracle.setPrice(values[i]);
+      const topOfPeriod = (await getTopOfPeriod()) + PERIOD;
+      await time.increaseTo(topOfPeriod);
+      await volOracle.mockCommit(
+        asset == WETH_ADDRESS ? ethusdcPool : wbtcusdcPool
```
kenchangh, line 2696: `asset === WETH_ADDRESS`

test/RibbonThetaVault.ts
```diff
@@ -316,17 +332,36 @@ function behavesLikeRibbonOptionsVault(params: {
...
+      );
+      const StrikeSelection = await getContractFactory(
+        "StrikeSelection",
+        ownerSigner
+      );
+
+      volOracle = await TestVolOracle.deploy(PERIOD);
+
+      optionsPremiumPricer = await OptionsPremiumPricer.deploy(
+        params.asset == WETH_ADDRESS ? ethusdcPool : wbtcusdcPool,
```
kenchangh, line 352: `params.asset === WETH_ADDRESS`

test/RibbonThetaVault.ts
```diff
@@ -316,17 +332,36 @@ function behavesLikeRibbonOptionsVault(params: {
...
+        volOracle.address,
+        params.asset == WETH_ADDRESS
```
kenchangh, line 354: `params.asset === WETH_ADDRESS`

test/StrikeSelection.ts
```diff
@@ -7,7 +7,7 @@ import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers";
 
 const { getContractFactory } = ethers;
 
-describe("StrikeSelection", () => {
+describe.skip("StrikeSelection", () => {
```
kenchangh, line 10: Why are we skipping this?

test/StrikeSelectionE2E.ts
```diff
@@ -0,0 +1,250 @@
...
+        underlyingPrice
+          .add(
+            deltaAtUnderlying
+              .sub(targetDelta)
+              .div(500)
+              .mul(100)
+              .mul(BigNumber.from(10).pow(await wethPriceOracle.decimals()))
+          )
+          .toString()
+      );
```
kenchangh, line 155: Do you mind helping clarify what this snippet does? Why do we need to `.div(500).mul(100)`

kenchangh: Okay I see now. You're just calculating the final strike price given the delta at underlying. Why don't we just hardcode the results in the test itself?

chuddster: we could do that. but just in case we wanted to customize the target delta within specific test we can do that (instead of once in beforeEach).

the .div(500) is because I found that the strike price moves by 100 (for ether) at approximately blocks of 0.05 in delta.

kenchangh (approved):

PR #27, opened 2021-06-26
management / performance fee

chuddster:


contracts/libraries/VaultLifecycle.sol
```diff
@@ -420,7 +420,6 @@ library VaultLifecycle {
         require(owner != address(0), "!owner");
         require(feeRecipient != address(0), "!feeRecipient");
         require(performanceFee > 0, "!performanceFee");
-        require(managementFee > 0, "!managementFee");
         require(bytes(tokenName).length > 0, "!tokenName");
```
chuddster, line 423: might not want to have a management fee initially

contracts/libraries/VaultLifecycle.sol
```diff
@@ -411,16 +411,15 @@ library VaultLifecycle {
     function verifyConstructorParams(
         address owner,
         address feeRecipient,
-        uint256 performanceFee,
         uint256 managementFee,
```
kenchangh, line 414: Since we're not checking `managementFee > 0` we can just remove it entirely right?

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -608,6 +649,46 @@ contract RibbonThetaVault is OptionsVaultStorage {
...
+    function _collectVaultFees(uint256 currentLockedBalance)
+        private
+        returns (uint256 vaultFee)
+    {
+        // Take management fee
+        if (managementFee > 0) {
+            uint256 managementFeeInAsset =
+                currentLockedBalance.mul(managementFee).div(100 * 10**6);
+            vaultFee = managementFeeInAsset;
+        }
```
kenchangh, line 666: We should also take management fee only when the vault is profitable for the week (just like the performance fee)

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -608,6 +649,46 @@ contract RibbonThetaVault is OptionsVaultStorage {
...
+                currentLockedBalance
+                    .sub(vaultState.totalPending)
+                    .sub(prevLockedAmount)
+                    .mul(performanceFee)
+                    .div(100 * 10**6);
+            vaultFee = vaultFee.add(performanceFeeInAsset);
+        }
+
+        if (vaultFee > 0) {
+            transferAsset(payable(feeRecipient), vaultFee);
```
kenchangh, line 688: We should emit an event with `(performanceFee, vaultFee, round)`

kenchangh (approved):

PR #28, opened 2021-06-28
add master cron job set up

chuddster:


scripts/masterCron.ts
```diff
@@ -0,0 +1,272 @@
...
+    async function () {
+      await masterJob();
+    },
+    null,
+    true,
+    "Atlantic/Reykjavik"
+  );
+
+  var volOracleJob = new CronJob(
+    new Date(CLOSEST_VALID_TIME),
```
chuddster, line 249: cron job within a cron job. first we wait until the first valid timestamp from now where we can call commit. then from that point on we do it every 12 hours, or however long the period is

scripts/masterCron.ts
```diff
@@ -0,0 +1,272 @@
...
+  // 2. wait an hour (timelock period)
+  await sleep(delayBeforeRoll[0]);
+  // 3. rollToNextOption
+  let auctionCounters = await runTX(
+    vaultArtifact.abi,
+    provider,
+    signer,
+    network,
+    "rollToNextOption"
+  );
```
kenchangh, line 178: What if we just created a few cron jobs instead of sleeping for an hour? Better to decouple them imo

scripts/masterCron.ts
```diff
@@ -0,0 +1,272 @@
...
+  const volOracleArtifact = await hre.artifacts.readArtifact("VolOracle");
+
+  const volOracle = new ethers.Contract(
+    VOL_ORACLE,
+    volOracleArtifact.abi,
+    provider
+  );
+
+  for (let univ3poolName in deployments[network].univ3pools) {
+    volOracle.commit(deployments[network].univ3pools[univ3poolName]);
```
kenchangh, line 210: missing await here

kenchangh (dismissed):
LGTM

scripts/masterCron.ts
```diff
@@ -204,6 +172,8 @@ async function settleAuction(auctionCounters: Array<number>) {
     network,
     "burnRemainingOTokens"
   );
+
+  auctionIDs = [];
```
kenchangh, line 176: Do we use this anywhere?

scripts/masterCron.ts
```diff
@@ -290,6 +262,7 @@ function run() {
         VOL_ORACLE_CRON,
         async function () {
           await updateVolatility();
+          console.log("hello");
```
kenchangh, line 265: Forgot to remove this? :)

scripts/masterCron.ts
chuddster: ye this is a global. rolltonextoption populates with id's. in settleAuction they are used to settle correct auctions. then they are cleared. repeat

chuddster: https://github.com/ribbon-finance/ribbon-v2/blob/b66275966278a0d7ac447a2afbc9c102ae12d1f6/scripts/masterCron.ts#L115

kenchangh: What if it fails for whatever reason? We will be losing the auction IDs needed to settle. We should persist it somewhere imo

scripts/masterCron.ts
chuddster: changed

scripts/masterCron.ts
kenchangh: Maybe use LevelDB or something? https://github.com/google/leveldb

kenchangh: Better yet - we can just expose this info on the contract itself. Then the cron script can lookup the active auctions.

kenchangh (approved):
Looks good to go 👍

PR #34, opened 2021-07-07
Master cron edit2

chuddster:


scripts/masterCron.ts
```diff
@@ -105,11 +108,12 @@ async function runTX(
...
     }
 
-    let gasPrice = await gas(network);
+    let newGasPrice = (await gas(network)).toString();
 
     try {
       const tx = await vault.connect(signer)[`${method}()`]({
-        gasPrice,
+        gasPrice: newGasPrice,
+        gasLimit: GAS_LIMIT,
```
kenchangh, line 116: Btw the GAS_LIMIT only applies for `commit()`, it's larger for other TV functions

chuddster: ah yes, edited

kenchangh (approved):

PR #35, opened 2021-07-09
Add delta vault commitAndClose

chuddster:


contracts/vaults/RibbonDeltaVault.sol
```diff
@@ -146,7 +146,36 @@ contract RibbonDeltaVault is RibbonVault, OptionsDeltaVaultStorage {
...
+        _closeLong(oldOption);
+    }
+
+    /**
+     * @notice Closes the existing long position for the vault.
+     */
+    function _closeLong(address oldOption) private {
+        optionState.currentOption = address(0);
+        vaultState.lastLockedAmount = vaultState.lockedAmount;
+        vaultState.lockedAmount = 0;
```
kenchangh, line 166: Does the `lockedAmount` variable even make sense in the context of Delta vault? We're not locking anything and spending it right away, so we *could* be double counting here

contracts/vaults/RibbonDeltaVault.sol
```diff
@@ -29,13 +29,13 @@ contract RibbonDeltaVault is RibbonVault, OptionsDeltaVaultStorage {
 
     event OpenLong(
         address indexed options,
-        uint256 depositAmount,
+        uint256 purchaseAmount,
```
kenchangh, line 32: Need to add `uint256 premium` also

contracts/vaults/RibbonDeltaVault.sol
chuddster: it doesn't _really_ make sense, but trying to abstract as much as possible. I can repurpose locked amount to be the premium paid during the week. used in RibbonVault parent in totalBalance() 

if you think its an unnecessary stretch can just keep locked lingo specific to theta vault and introduce something new

kenchangh: Gotcha, thanks for the explanation

contracts/vaults/RibbonDeltaVault.sol
```diff
@@ -147,14 +148,12 @@ contract RibbonDeltaVault is RibbonVault, OptionsDeltaVaultStorage {
...
      */
     function commitAndClose() external onlyOwner nonReentrant {
-        address oldOption = optionState.currentOption;
+      address oldOption = optionState.currentOption;
 
-        optionState.nextOption = counterpartyThetaVault
-            .optionState()
-            .nextOption;
-        optionState.nextOptionReadyAt = uint32(block.timestamp.add(delay));
+      optionState.nextOption = counterpartyThetaVault.optionState().nextOption;
```
kenchangh: `require(counterpartyThetaVault.optionState().nextOption != address(0)`

contracts/vaults/RibbonDeltaVault.sol
```diff
@@ -146,7 +147,36 @@ contract RibbonDeltaVault is RibbonVault, OptionsDeltaVaultStorage {
...
      *         This allows all the users to withdraw if the next option is malicious.
      */
-    function commitAndClose() external onlyOwner nonReentrant {}
+    function commitAndClose() external onlyOwner nonReentrant {
+        address oldOption = optionState.currentOption;
+
+        optionState.nextOption = counterpartyThetaVault
+            .optionState()
+            .nextOption;
+        optionState.nextOptionReadyAt = uint32(block.timestamp.add(delay));
```
kenchangh, line 156: We could streamline the state updates by stuffing these into `_closeLong` instead

kenchangh (approved):

PR #36, opened 2021-07-09
update voloracle address

chuddster:


kenchangh (dismissed):

scripts/masterCron.ts
```diff
@@ -268,6 +268,7 @@ async function run() {
         "Atlantic/Reykjavik"
       );
 
+      await updateVolatility();
       _.start();
```
kenchangh, line 272: Do we want to kick off the next cronjob first, before running the current one?

kenchangh (approved):

PR #37, opened 2021-07-09
Delta vault roll to next option

chuddster:


contracts/libraries/GnosisAuction.sol
```diff
@@ -103,6 +118,58 @@ library GnosisAuction {
...
+    function placeBid(BidDetails memory bidDetails)
+        internal
+        returns (uint256 oTokensToBuy, uint256 bidAmount)
+    {
+        // calculate how much to allocate
+        bidAmount = bidDetails
+            .lockedBalance
+            .mul(bidDetails.optionAllocationPct)
+            .div(100 * 10**2);
+        oTokensToBuy = 0; //bidAmount.div(PREMIUM)
```
chuddster, line 130: still TBD what premium we are willing to give per oToken. 

should we offer the floor premium of the auction?

contracts/libraries/GnosisAuction.sol
```diff
@@ -103,6 +120,63 @@ library GnosisAuction {
...
 
+    function placeBid(BidDetails memory bidDetails)
+        internal
+        returns (uint256 oTokensToBuy, uint256 bidAmount)
+    {
+        // calculate how much to allocate
+        bidAmount = bidDetails
+            .lockedBalance
+            .mul(bidDetails.optionAllocationPct)
+            .div(100 * 10**2);
```
kenchangh, line 131: Why do we have to do `100 * 10**2`?

contracts/libraries/GnosisAuction.sol
```diff
@@ -20,6 +20,12 @@ library GnosisAuction {
         address manager
     );
 
+    event PlaceAuctionBid(
+        uint256 auctionId,
+        address auctioningToken,
+        uint256 bidAmount
```
kenchangh, line 26: Let's add `address bidder`

contracts/interfaces/IGnosisAuction.sol
```diff
@@ -68,4 +68,9 @@ interface IGnosisAuction {
         bytes32[] memory _prevSellOrders,
         bytes calldata allowListCallData
     ) external returns (uint64);
+
+    function claimFromParticipantOrder(
```
chuddster, line 72: bidder actually needs to call [claimFromParticipantOrder](https://github.com/gnosis/ido-contracts/blob/a766a8b30de14781961158b772393cc49b900714/contracts/EasyAuction.sol#L568) as well. so will need to call this on behalf of delta vault in cron job in separate pr

contracts/libraries/GnosisAuction.sol
```diff
@@ -103,6 +123,73 @@ library GnosisAuction {
...
+        );
+
+        // divide the `asset` sellAmount by the target premium per oToken to
+        // get the number of oTokens to buy (8 decimals)
+        buyAmount = uint96(
+            uint256(sellAmount)
+                .mul(10**bidDetails.assetDecimals)
+                .div(bidDetails.optionPremium)
+                .mul(10**8)
+                .div(10**bidDetails.assetDecimals)
```
chuddster, line 148: multiplying by asset decimals twice - once for bidDetails.optionPremium which has `asset` decimals and once for shifting it to 8 decimals

contracts/vaults/RibbonDeltaVault.sol
```diff
@@ -199,6 +200,7 @@ contract RibbonDeltaVault is RibbonVault, OptionsDeltaVaultStorage {
         bidDetails.lockedBalance = lockedBalance;
         bidDetails.optionAllocationPct = optionAllocationPct;
         bidDetails.optionPremium = optionPremium;
+        bidDetails.bidder = owner();
```
kenchangh, line 206: This should be `bidDetails.bidder = msg.sender` to save gas

contracts/interfaces/IGnosisAuction.sol
kenchangh: Do we want to make a pass-through function on the Delta Vault that calls `claimFromParticipantOrder`? Feels better to have everything on 1 contract so people can poke it

contracts/libraries/GnosisAuction.sol
```diff
@@ -124,47 +125,55 @@ library GnosisAuction {
...
     {
         // calculate how much to allocate
-        bidAmount = bidDetails
-            .lockedBalance
-            .mul(bidDetails.optionAllocationPct)
-            .div(10000);
+        sellAmount = uint96(
+            bidDetails.lockedBalance.mul(bidDetails.optionAllocationPct).div(
+                10000
+            )
```
kenchangh, line 138: Check for uint96 overflow

contracts/libraries/GnosisAuction.sol
```diff
@@ -124,47 +125,55 @@ library GnosisAuction {
...
-            bidAmount <= type(uint96).max,
-            "bidAmount > type(uint96) max value!"
+            sellAmount <= type(uint96).max,
+            "sellAmount > type(uint96) max value!"
         );
         require(
-            oTokensToBuy <= type(uint96).max,
-            "oTokensToBuy > type(uint96) max value!"
+            buyAmount <= type(uint96).max,
+            "buyAmount > type(uint96) max value!"
```
kenchangh: Checking for uint96 overflow here doesn't actually do anything. We need to do the checks before casting it to uint96.

This works:

```solidity
uint256 number = someNumber;
require(number <= type(uint96.max));
uint96 smallNumber = uint96(number);
```

This does not work:

```solidity
uint256 number = someNumber;
// casting already truncates the integer to be the size uint96.
uint96 smallNumber = uint96(number); 
require(smallNumber <= type(uint96.max));
```

chuddster: yes good point

contracts/vaults/RibbonDeltaVault.sol
```diff
@@ -152,13 +161,12 @@ contract RibbonDeltaVault is RibbonVault, OptionsDeltaVaultStorage {
 
         address counterpartyNextOption =
             counterpartyThetaVault.optionState().nextOption;
-        require(counterpartyNextOption != address(0));
+        require(counterpartyNextOption != address(0), "!thetavaultclosed");
```
kenchangh, line 164: nit deltavaultclosed

kenchangh (approved):

PR #38, opened 2021-07-09
V1 -> V2 migration

Gadget212:
V2 changes to enable migration from V1 vaults. Notable side effects are a configurable `initialSharePrice` and a new `depositFor` pattern. The latter allows vaults to deposit on behalf of users despite the time delay on share distribution.


contracts/libraries/ShareMath.sol
```diff
@@ -53,7 +53,8 @@ library ShareMath {
         Vault.DepositReceipt memory depositReceipt,
         uint16 currentRound,
         uint256 pps,
-        uint8 decimals
+        uint8 decimals,
+        uint256 initialSharePrice
```
kenchangh, line 57: I don't see the `initialSharePrice` variable used?

contracts/libraries/ShareMath.sol
```diff
@@ -71,7 +72,7 @@ library ShareMath {
 
             unredeemedShares = uint128(unredeemedShares256);
         } else {
-            unredeemedShares = depositReceipt.unredeemedShares;
+            unredeemedShares = defaultSharePrice.mul(depositReceipt.unredeemedShares);
```
kenchangh, line 75: Needs to rename to `initialSharePrice`

kenchangh: Think we have to retain the old statement:

```solidity
unredeemedShares = depositReceipt.unredeemedShares;
```

Instead, we should add another condition that checks that it's the first round:

```solidity
} else if (depositReceipt.round == 1) {
  unredeemedShares = initiatalSharePrice.mul(depositReceipt.unredeemedShares);
} else {
  unredeemedShares = depositReceipt.unredeemedShares;
}
```

Gadget212: I original misunderstood the logic in getSharePrice, and have moved the initialSharePrice adjustment to the rollover logic in `vaultLifecycle.sol`, where it's used when currentSupply is 0.

contracts/libraries/VaultLifecycle.sol
```diff
@@ -142,7 +142,7 @@ library VaultLifecycle {
 
         newPricePerShare = currentSupply > 0
             ? singleShare.mul(roundStartBalance).div(currentSupply)
-            : singleShare;
+            : vaultParams.initialSharePrice;
```
kenchangh, line 145: I like this approach because it moves the storage access of `vaultParams.initialSharePrice` to `rollToNextOption`, which means it won't be in the deposit/withdrawal flow

contracts/vaults/base/RibbonVault.sol
```diff
@@ -251,17 +255,18 @@ contract RibbonVault is OptionsVaultStorage {
...
+            depositReceipts[creditor];
 
         // If we have an unprocessed pending deposit from the previous rounds, we have to process it.
         uint128 unredeemedShares =
             depositReceipt.getSharesFromReceipt(
                 currentRound,
                 roundPricePerShare[depositReceipt.round],
-                vaultParams.decimals
+                vaultParams.decimals,
+                vaultParams.initialSharePrice
```
kenchangh, line 269: We can remove the `initialSharePrice` param entirely since deposits on the first round should not affect share price

contracts/vaults/base/RibbonVault.sol
```diff
@@ -325,7 +330,8 @@ contract RibbonVault is OptionsVaultStorage {
...
```
kenchangh, line 334: Can remove https://github.com/ribbon-finance/ribbon-v2/pull/38/files#r668738306

contracts/vaults/base/RibbonVault.sol
```diff
@@ -574,7 +580,8 @@ contract RibbonVault is OptionsVaultStorage {
             depositReceipt.getSharesFromReceipt(
                 vaultState.round,
                 roundPricePerShare[depositReceipt.round],
-                vaultParams.decimals
+                vaultParams.decimals,
+                vaultParams.initialSharePrice
```
kenchangh, line 584: Can remove this https://github.com/ribbon-finance/ribbon-v2/pull/38/files#r668738306

kenchangh (dismissed):
Looks good to me, let's fix the test and get it merged

kenchangh (approved):

PR #40, opened 2021-07-12
Add mint amount to OpenShort event

kenchangh:
This event parameter was missing from v1, which was causing pain


kenchangh:
Stale, revisiting later

PR #42, opened 2021-07-15
add claimFromParticipantOrder

chuddster:


scripts/masterCron.ts
```diff
@@ -81,12 +84,27 @@ async function settleAuctions(
...
       });
+
       await log(`GnosisAuction-settleAuction()-${auctionID}: ${tx.hash}`);
     } catch (error) {
       await log(
         `@everyone GnosisAuction-settleAuction()-${auctionID}: failed with error ${error}`
       );
     }
+
+    try {
```
chuddster, line 95: better to separate _claimFromParticipantOrder_ into another method?

kenchangh: Looks fine how we're doing it now

kenchangh (approved):

PR #43, opened 2021-07-15
add delta vault instant withdrawal any time

chuddster:


contracts/vaults/base/RibbonVault.sol
```diff
@@ -347,15 +347,20 @@ contract RibbonVault is OptionsVaultStorage {
...
 
         uint16 currentRound = vaultState.round;
         require(amount > 0, "!amount");
-        require(!depositReceipt.processed, "Processed");
-        require(depositReceipt.round == currentRound, "Invalid round");
+
+        if (isThetaVault) {
+            require(!depositReceipt.processed, "Processed");
+            require(depositReceipt.round == currentRound, "Invalid round");
+        }
```
kenchangh, line 362: What's the rationale behind doing these checks only for Theta Vault? Isn't the withdraw instantly completely different for Delta Vault? Like we don't even have to concern ourselves with deposit receipt stuff

chuddster: isnt deposit receipt still relevant for how deposits work? withdrawals I agree.

contracts/vaults/RibbonDeltaVault.sol
```diff
@@ -233,5 +280,46 @@ contract RibbonDeltaVault is RibbonVault, OptionsDeltaVaultStorage {
...
+     * @return the amount left to withdraw
+     */
+    function _withdrawFromNewDeposit(uint256 amount) private returns (uint256) {
+        Vault.DepositReceipt storage depositReceipt =
+            depositReceipts[msg.sender];
+
+        uint256 receiptAmount = depositReceipt.amount;
+
+        // Immediately get what is in the pending deposits, without need for checking pps
+        if (depositReceipt.round == vaultState.round && receiptAmount > 0) {
```
chuddster, line 314: we dont need to do an additional check on whether deposit receipt has been processed right. processed is only set to true in redeem. but the condition in redeem is that the round of the deposit is before the current round. but here we are making sure the deposit round is the same as current round. so it would not be possible for it to be processed.

contracts/vaults/RibbonDeltaVault.sol
```diff
@@ -233,5 +280,46 @@ contract RibbonDeltaVault is RibbonVault, OptionsDeltaVaultStorage {
...
+
+        _updatePPS();
+    }
+
+    /**
+     * @notice Updates the price per share of the current round. The current round
+     * pps will change right after call rollToNextOption as the gnosis auction contract
+     * takes custody of a % of `asset` tokens, and right after we claim the tokens from
+     * the action as we may recieve some of `asset` tokens back alongside the oToken,
+     * depending on the gnosis auction outcome. Finally it will change at the end of the week
```
chuddster, line 292: rollToNextOption: calling VaultLifecycle.placeBid gives gnosis contract our funds we approved. pps changes for some time here. 

claimParticipantOrder: every leaf in the if/else tree its possible for some amount of bidding token to be sent back to us. mainly in [this](https://github.com/gnosis/ido-contracts/blob/a766a8b30de14781961158b772393cc49b900714/contracts/EasyAuction.sol#L613) case if settle price comes out to some number where we get some of `asset` back. pps changes for some time here.

contracts/vaults/RibbonDeltaVault.sol
```diff
@@ -233,5 +280,46 @@ contract RibbonDeltaVault is RibbonVault, OptionsDeltaVaultStorage {
...
+     * if the oTokens are ITM
+     */
+    function _updatePPS() private {
+        (, uint256 newPricePerShare, ) =
+            VaultLifecycle.rollover(totalSupply(), vaultParams, vaultState);
```
kenchangh, line 297: There's a lot of extra steps that we can avoid in the `rollover` function, it's worth it to just copy paste the newPricePerShare calculation into a new helper function

contracts/vaults/RibbonDeltaVault.sol
```diff
@@ -214,6 +259,8 @@ contract RibbonDeltaVault is RibbonVault, OptionsDeltaVaultStorage {
         auctionSellOrder.buyAmount = uint96(buyAmount);
         auctionSellOrder.userId = userId;
 
+        _updatePPS();
```
kenchangh, line 262: Good use for a modifier

contracts/vaults/RibbonDeltaVault.sol
kenchangh: Yea we don't have to check it

contracts/vaults/RibbonDeltaVault.sol
```diff
@@ -148,6 +151,48 @@ contract RibbonDeltaVault is RibbonVault, OptionsDeltaVaultStorage {
...
+
+            require(sharesLeft <= totalShares, "Insufficient balance");
+
+            // Burn both redeemed shares and custodied shares of user
+            if (heldByAccount > 0) {
+                _burn(msg.sender, heldByAccount);
+            }
+            if (heldByVault > 0) {
+                _burn(address(this), heldByVault);
+            }
```
kenchangh, line 188: Based on ChainSafe's comments, we should avoid using this pattern of burning from both the vault and account on Theta Vault. I tend to agree.

One way to refactor this is to force a maxRedeem to happen before we burn, so we burn the shares from the user's account instead

contracts/vaults/RibbonDeltaVault.sol
chuddster: agree

contracts/vaults/RibbonDeltaVault.sol
```diff
@@ -122,6 +125,41 @@ contract RibbonDeltaVault is RibbonVault, OptionsDeltaVaultStorage {
...
+     * if the oTokens are ITM
+     */
+    modifier updatePPS(bool isWithdraw) {
+        if (!isWithdraw) {
+            _;
+        }
+
+        if (
+            !isWithdraw ||
+            roundPricePerShare[vaultState.round] <= PLACEHOLDER_UINT
```
chuddster, line 143: I included case where we are doing deposit/withdraw sequence for very first time in vault. we will have to update price per share in this case. but we will have to do this check for every withdraw afterwards (because rollToNextOption will be called)

kenchangh (approved):
Looks good to me

PR #44, opened 2021-07-15
Ken/chainsafe fix3

kenchangh:


contracts/libraries/VaultLifecycle.sol
```diff
@@ -436,24 +436,28 @@ library VaultLifecycle {
...
+        uint256 friday8am =
+            nextFriday - (nextFriday % (60 * 60 * 24)) + (8 * 60 * 60);
 
-        uint256 friday;
-        if (dayOfWeek > 5) {
-            friday = nextWeek - 86400 * (dayOfWeek - 5);
-        } else {
-            friday = nextWeek + 86400 * (5 - dayOfWeek);
+        // If the passed currentExpiry is day=Friday hour>8am, we simply increment it by a week to next Friday
+        if (currentExpiry >= friday8am) {
```
chuddster, line 458: this is only if currentExpiry is past friday 8am but before that week's sunday?

chuddster (approved):

contracts/libraries/VaultLifecycle.sol
kenchangh: Yep, this scenario only happens on Friday

PR #45, opened 2021-07-16
Ken/cs refactor

kenchangh:
- Force an implicit max redeem whenever users call `initiateWithdraw`. This simplifies the math we have to do in `initiateWithdraw`
- Call `_burn` when we do `completeWithdraw`

### DIFF
--- a/contracts/vaults/RibbonThetaVault.sol
+++ b/contracts/vaults/RibbonThetaVault.sol
@@ -18,6 +18,7 @@ import {
     IStrikeSelection,
     IOptionsPremiumPricer
 } from "../interfaces/IRibbon.sol";
+import "hardhat/console.sol";
 
 contract RibbonThetaVault is OptionsVaultStorage {
     using SafeERC20 for IERC20;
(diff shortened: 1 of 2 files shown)


contracts/vaults/RibbonThetaVault.sol
```diff
@@ -424,20 +424,21 @@ contract RibbonThetaVault is OptionsVaultStorage {
...
 
         bool topup = withdrawal.initiated && withdrawal.round == currentRound;
 
-        (uint256 heldByAccount, uint256 heldByVault) =
-            shareBalances(msg.sender);
-
-        uint256 vaultRemainder = heldByVault.sub(withdrawal.shares);
-        uint256 totalShares = heldByAccount.add(vaultRemainder);
-
-        require(shares <= totalShares, "Insufficient balance");
```
chuddster, line 439: why do we remove this require?

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -458,12 +459,7 @@ contract RibbonThetaVault is OptionsVaultStorage {
             uint256(vaultState.queuedWithdrawShares).add(shares)
         );
 
-        // We need to debit the user's account when they are trying to withdraw
-        // more than what's available in the vault, accounting for previous withdrawals
-        if (shares > vaultRemainder) {
-            uint256 debitShares = uint256(shares).sub(vaultRemainder);
-            _transfer(msg.sender, address(this), debitShares);
-        }
+        _transfer(msg.sender, address(this), shares);
```
chuddster, line 462: clever

contracts/vaults/RibbonThetaVault.sol
```diff
@@ -424,20 +424,21 @@ contract RibbonThetaVault is OptionsVaultStorage {
     function initiateWithdraw(uint128 shares) external nonReentrant {
         require(shares > 0, "!shares");
 
+        // We do a max redeem before initiating a withdrawal
+        // But we check if they must first have unredeemed shares
+        if (
+            depositReceipts[msg.sender].amount > 0 ||
+            depositReceipts[msg.sender].unredeemedShares > 0
+        ) {
+            _redeem(0, true);
```
chuddster, line 433: in delta vault I also check if shares > heldByAccount to save gas

chuddster (approved):

PR #47, opened 2021-07-19
update yearn vault to v2

chuddster:


contracts/libraries/VaultLifecycleYearn.sol
```diff
@@ -0,0 +1,554 @@
...
+            address otokenAddress,
+            uint256 premium,
+            uint256 strikePrice,
+            uint256 delta
+        )
+    {
+        uint256 expiry;
+
+        // uninitialized state
+        if (closeParams.currentOption <= address(1)) {
```
kenchangh, line 56: I [changed this recently](https://github.com/ribbon-finance/ribbon-v2/blob/chainsafe-audit/contracts/libraries/VaultLifecycle.sol#L51), needs to be `if (closeParams.currentOption == address(0)) {`

contracts/vaults/YearnVaults/RibbonDeltaYearnVault.sol
```diff
@@ -0,0 +1,357 @@
...
+import {
+    OptionsDeltaYearnVaultStorage
+} from "../../storage/OptionsVaultYearnStorage.sol";
+import {Vault} from "../../libraries/Vault.sol";
+import {VaultLifecycleYearn} from "../../libraries/VaultLifecycleYearn.sol";
+import {ShareMath} from "../../libraries/ShareMath.sol";
+import {RibbonVault} from "./base/RibbonVault.sol";
+import {IRibbonThetaVault} from "../../interfaces/IRibbonThetaVault.sol";
+
+contract RibbonDeltaYearnVault is RibbonVault, OptionsDeltaYearnVaultStorage {
```
kenchangh, line 18: Since the Delta vaults don't deal with yvUSDC, why do we need to have a Yearn implementation?

contracts/vaults/YearnVaults/RibbonThetaYearnVault.sol
```diff
@@ -0,0 +1,350 @@
...
+     * @param _marginPool is the contract address for providing collateral to opyn
+     * @param _gnosisEasyAuction is the contract address that facilitates gnosis auctions
+     */
+    constructor(
+        address _weth,
+        address _usdc,
+        address _oTokenFactory,
+        address _gammaController,
+        address _marginPool,
+        address _gnosisEasyAuction
```
kenchangh, line 89: We're missing passing the Yearn registry

contracts/vaults/YearnVaults/base/RibbonVault.sol
```diff
@@ -0,0 +1,815 @@
...
+            uint256 managementFeeInAsset =
+                managementFee > 0
+                    ? currentLockedBalance.mul(managementFee).div(100 * 10**6)
+                    : 0;
+
+            vaultFee = performanceFeeInAsset.add(managementFeeInAsset);
+        }
+
+        if (vaultFee > 0) {
+            transferAsset(payable(feeRecipient), vaultFee);
```
kenchangh, line 673: Following this function here, just confirming that we're going to collect fees in yvUSDC?

kenchangh: I think it's fine to collect fees in yvUSDC btw, if it simplifies things

contracts/vaults/YearnVaults/base/RibbonVault.sol
```diff
@@ -0,0 +1,815 @@
...
+     * @return lockedBalance is the new balance used to calculate next option purchase size or collateral size
+     */
+    function _rollToNextOption() internal returns (address, uint256) {
+        require(block.timestamp >= optionState.nextOptionReadyAt, "Not ready");
+
+        address newOption = optionState.nextOption;
+        require(newOption != address(0), "!nextOption");
+
+        // Wrap entire `asset` balance to `collateralToken` balance
+        _wrapToYieldToken();
```
kenchangh, line 609: Is it possible to move this to the bottom of the function? We're following the [check-effect-interaction pattern](https://fravoll.github.io/solidity-patterns/checks_effects_interactions.html) so we should do interactions that involve asset transfers last.

contracts/vaults/YearnVaults/RibbonThetaYearnVault.sol
```diff
@@ -0,0 +1,350 @@
...
+
+        optionAuctionID = VaultLifecycleYearn.startAuction(auctionDetails);
+    }
+
+    /**
+     * @notice Burn the remaining oTokens left over from gnosis auction.
+     */
+    function burnRemainingOTokens() external onlyOwner nonReentrant {
+        // Wrap entire `asset` balance to `collateralToken` balance
+        _wrapToYieldToken();
```
kenchangh, line 322: Same here https://github.com/ribbon-finance/ribbon-v2/pull/47/files#r673360238

contracts/vaults/YearnVaults/RibbonDeltaYearnVault.sol
chuddster: I was thinking buying options whilst keeping the reserves in yvUSDC so your net loss every week is smaller. But I guess not worth.

contracts/vaults/YearnVaults/RibbonThetaYearnVault.sol
chuddster: in v1, we had it in constructor because the `asset` parameter was there. but the `asset` parameter is now in initialize, so I pass it in there. `yearnRegistry` is only used once to get the collateral token, so I don't think it is necessary to pass it in constructor, hold it as a constant, and then reference it in initialize. Just pass it into initialize

chuddster: https://github.com/ribbon-finance/ribbon-v2/pull/47/files#diff-93cfb819847795b1c4f019da7b67280790d8b43a53485ac94b47b96d684b3e20R159

contracts/vaults/YearnVaults/base/RibbonVault.sol
chuddster: I changed it to be yvUSDC

contracts/vaults/YearnVaults/RibbonThetaYearnVault.sol
kenchangh: Understood, sounds good

contracts/vaults/YearnVaults/base/RibbonVault.sol
```diff
@@ -412,21 +415,24 @@ contract RibbonVault is OptionsVaultYearnStorage, DSMath {
...
+     * @param recipient is the recipient
      * @param amount is the withdraw amount in `asset`
      * @return withdrawAmount is the withdraw amount in `collateralToken`
      */
-    function _withdrawYieldAndBaseToken(uint256 amount)
-        internal
-        returns (uint256 withdrawAmount)
-    {
+    function _withdrawYieldAndBaseToken(
+        address payable recipient,
```
kenchangh, line 423: We don't need `payable`

contracts/vaults/YearnVaults/base/RibbonVault.sol
```diff
@@ -600,6 +601,28 @@ contract RibbonVault is OptionsVaultYearnStorage, DSMath {
         }
     }
 
+    /*
+      Upgrades the vault to point to the latest yearn vault for the asset token
+    */
+    function upgradeYearnVault() external onlyOwner {
```
chuddster, line 607: this should be called in between commitAndClose() and rollToNextOption(). only time we have all the collateral back in the vault

contracts/vaults/YearnVaults/base/RibbonVault.sol
```diff
@@ -600,6 +601,28 @@ contract RibbonVault is OptionsVaultYearnStorage, DSMath {
...
+            YEARN_WITHDRAWAL_SLIPPAGE
+        );
+        _upgradeYearnVault();
+    }
+
+    function _upgradeYearnVault() internal {
+        address collateralAddr =
+            IYearnRegistry(YEARN_REGISTRY).latestVault(vaultParams.asset);
+        require(collateralAddr != address(0), "!collateralToken");
+        collateralToken = IYearnVault(collateralAddr);
```
kenchangh, line 623: Needs to deposit into the new vault

contracts/vaults/YearnVaults/RibbonThetaYearnVault.sol
```diff
@@ -0,0 +1,379 @@
...
+
+        vaultState.lockedAmount = uint104(lockedBalance);
+
+        emit OpenShort(newOption, lockedBalance, msg.sender);
+
+        VaultLifecycleYearn.createShort(
+            GAMMA_CONTROLLER,
+            MARGIN_POOL,
+            newOption,
+            wdiv(
```
kenchangh, line 310: We can try removing `DSMath` entirely as an import by copy pasting the wdiv/wmul functions into the contract, see if that helps with the size

contracts/vaults/YearnVaults/RibbonThetaYearnVault.sol
```diff
@@ -0,0 +1,379 @@
...
+import {GnosisAuction} from "../../libraries/GnosisAuction.sol";
+import {
+    OptionsThetaYearnVaultStorage
+} from "../../storage/OptionsVaultYearnStorage.sol";
+import {Vault} from "../../libraries/Vault.sol";
+import {VaultLifecycleYearn} from "../../libraries/VaultLifecycleYearn.sol";
+import {ShareMath} from "../../libraries/ShareMath.sol";
+import {RibbonVault} from "./base/RibbonVault.sol";
+
+contract RibbonThetaYearnVault is RibbonVault, OptionsThetaYearnVaultStorage {
```
kenchangh, line 17: We can remove `OptionsThetaYearnVaultStorage` since it's already inherited by RibbonVault?

contracts/vaults/YearnVaults/RibbonThetaYearnVault.sol
```diff
@@ -0,0 +1,379 @@
...
+
+    event PremiumDiscountSet(
+        uint256 premiumDiscount,
+        uint256 newPremiumDiscount
+    );
+
+    event AuctionDurationSet(
+        uint256 auctionDuration,
+        uint256 newAuctionDuration
+    );
```
kenchangh, line 55: We can remove these events imo

contracts/vaults/BaseVaults/RibbonThetaVault.sol
```diff
@@ -6,11 +6,11 @@ import {SafeMath} from "@openzeppelin/contracts/math/SafeMath.sol";
...
-import {VaultLifecycle} from "../libraries/VaultLifecycle.sol";
-import {ShareMath} from "../libraries/ShareMath.sol";
+import {GnosisAuction} from "../../libraries/GnosisAuction.sol";
+import {OptionsThetaVaultStorage} from "../../storage/OptionsVaultStorage.sol";
+import {Vault} from "../../libraries/Vault.sol";
+import {VaultLifecycle} from "../../libraries/VaultLifecycle.sol";
+import {ShareMath} from "../../libraries/ShareMath.sol";
 import {RibbonVault} from "./base/RibbonVault.sol";
 
 contract RibbonThetaVault is RibbonVault, OptionsThetaVaultStorage {
```
kenchangh, line 16: Same here https://github.com/ribbon-finance/ribbon-v2/pull/47/files#r676741657

contracts/vaults/YearnVaults/base/RibbonVault.sol
```diff
@@ -0,0 +1,744 @@
...
+        uint256 amount,
+        uint256 shares
+    );
+
+    event InitiateWithdraw(address account, uint256 shares, uint16 round);
+
+    event CollectVaultFees(
+        uint256 performanceFee,
+        uint256 vaultFee,
+        uint256 round
```
kenchangh, line 92: I must have missed this before, we should replace the `vaultFee` for `managementFee`

contracts/vaults/YearnVaults/base/RibbonVault.sol
```diff
@@ -0,0 +1,744 @@
...
+
+        if (vaultFee > 0) {
+            VaultLifecycleYearn.withdrawYieldAndBaseToken(
+                WETH,
+                vaultParams.asset,
+                address(collateralToken),
+                feeRecipient,
+                vaultFee
+            );
+            emit CollectVaultFees(performanceFee, vaultFee, vaultState.round);
```
kenchangh, line 600: Refer to https://github.com/ribbon-finance/ribbon-v2/pull/47/files#r676746116

`emit CollectVaultFees(performanceFee, managementFee, vaultState.round);`

contracts/vaults/YearnVaults/RibbonThetaYearnVault.sol
chuddster: good catch. originally made sense because we had theta / delta vaults so they diverged in whats needed for storage, but we have deleted delta

contracts/vaults/BaseVaults/RibbonThetaVault.sol
chuddster: we cannot do this because theta vaults inherit OptionsThetaVaultStorageV1 and delta vaults inherit OptionsDeltaVaultStorageV1.

both with different storage variables.

ribbon vault inherits OptionsVaultStorage which is the commonality between the two vaults

contracts/vaults/YearnVaults/base/RibbonVault.sol
chuddster: I actually added managementFee as well. 

vaultFee is how much we actually made that week which is worth passing in

contracts/vaults/YearnVaults/base/RibbonVault.sol
chuddster: see above

contracts/vaults/YearnVaults/base/RibbonVault.sol
kenchangh: The dev can just do `performanceFee + managementFee = vaultFee` so it's redundant?

test/RibbonThetaYearnVault.ts
```diff
@@ -0,0 +1,2964 @@
...
+
+    const getCurrentOptionExpiry = async () => {
+      const currentOption = await vault.currentOption();
+      const otoken = await getContractAt("IOtoken", currentOption);
+      return otoken.expiryTimestamp();
+    };
+
+    before(async function () {
+      // Reset block
+      await network.provider.request({
```
kenchangh, line 285: Is there some way to avoid doing `hardhat_reset`? Calling reset breaks the test coverage tool.

Alternatively, we can use [this solution to fork the state](https://github.com/sc-forks/solidity-coverage/issues/574#issuecomment-856330142).

test/RibbonThetaYearnVault.ts
```diff
@@ -0,0 +1,2964 @@
...
+        await depositIntoVault(params.depositAsset, vault, depositAmount);
+        const res = await vault
+          .connect(ownerSigner)
+          .commitAndClose({ from: owner });
+
+        const receipt = await res.wait();
+        assert.isAtMost(receipt.gasUsed.toNumber(), 1045000);
+        // console.log("commitAndClose", receipt.gasUsed.toNumber());
+      });
+    });
```
kenchangh, line 1297: Nit missing newline

contracts/libraries/VaultLifecycleYearn.sol
```diff
@@ -0,0 +1,793 @@
...
+            collateralAsset,
+            closeParams.USDC,
+            closeParams.delay
+        );
+
+        premium = GnosisAuction.getOTokenPremium(
+            otokenAddress,
+            optionsPremiumPricer,
+            premiumDiscount
+        );
```
kenchangh, line 98: Needs to account for the fact that the put contracts are collateralized with yvUSDC

contracts/vaults/YearnVaults/base/RibbonVault.sol
```diff
@@ -0,0 +1,672 @@
...
+    function _rollToNextOption() internal returns (address, uint256) {
+        require(block.timestamp >= optionState.nextOptionReadyAt, "!ready");
+
+        address newOption = optionState.nextOption;
+        require(newOption != address(0), "!nextOption");
+
+        (uint256 lockedBalance, uint256 newPricePerShare, uint256 mintShares) =
+            VaultLifecycleYearn.rollover(
+                totalSupply(),
+                totalBalance(),
```
kenchangh, line 464: I think we're meant to pass in the `collateralToken.balanceOf(address(this))` (yvUSDC balance) instead of `totalBalance()` (USDC balance)

This is likely why it's throwing off the calculation in `VaultLifecycleYearn.createShort`

test/RibbonThetaYearnVault.ts
```diff
@@ -110,13 +110,13 @@ describe("RibbonThetaYearnVault", () => {
...
     initialSharePrice: BigNumber.from("10").pow("6").toString(),
-    expectedMintAmount: BigNumber.from("52379566383"),
+    expectedMintAmount: BigNumber.from("52379588695"),
     auctionDuration: 21600,
     tokenDecimals: 6,
     isPut: true,
     gasLimits: {
       depositWorstCase: 151608,
-      depositBestCase: 130594,
+      depositBestCase: 130681,
```
kenchangh: Follow up PR after this, let's gas golf the deposit flow for ThetaYearnVault

kenchangh (approved):

PR #50, opened 2021-07-26
Cs fix 2

kenchangh:


contracts/libraries/GnosisAuction.sol
```diff
@@ -36,18 +36,6 @@ library GnosisAuction {
                 auctionDetails.gnosisEasyAuction
             );
 
-        if (
```
chuddster, line 39: why do we remove this?

contracts/libraries/VaultLifecycle.sol
```diff
@@ -178,16 +157,13 @@ library VaultLifecycle {
             (controller.getAccountVaultCounter(address(this))).add(1);
 
         IOtoken oToken = IOtoken(oTokenAddress);
-        uint256 strikePrice = oToken.strikePrice();
-        bool isPut = oToken.isPut();
```
chuddster, line 182: we are we removing these variables?

contracts/utils/StrikeSelection.sol
```diff
@@ -77,7 +77,10 @@ contract StrikeSelection is DSMath, Ownable {
         //   with certain margin of error
         //        return strike price
 
-        uint256 strike = assetPrice.sub(assetPrice % step);
+        uint256 strike =
+            isPut
+                ? assetPrice.sub(assetPrice % step)
+                : assetPrice.add(step - (assetPrice % step));
```
chuddster, line 83: why do we do this?

chuddster (commented):
Looks good for the most part. Added some comments

does this pr address all the issues in LINE BY LINE REVIEW or just some?

for example cant find:

RibbonThetaVault, line 430: Critical, `heldByVault` is not decreased after
withdrawal allowing users to withdraw the same `unredeemedShares` multiple
times.

contracts/utils/StrikeSelection.sol
kenchangh: 4.36. StrikeSelection, line 80: Note, `strike` should start with increased price in case of
a call.

contracts/libraries/VaultLifecycle.sol
kenchangh: Getting stack too deep error and it's only used once so I just called it directly below `otoken.isPut()`

contracts/libraries/GnosisAuction.sol
kenchangh: It's not needed because in most cases we're only approving the auction contract once, and we can trust that the auction will not overdraw the allowance

kenchangh:
> Looks good for the most part. Added some comments
> 
> does this pr address all the issues in LINE BY LINE REVIEW or just some?
> 
> for example cant find:
> 
> RibbonThetaVault, line 430: Critical, `heldByVault` is not decreased after
> withdrawal allowing users to withdraw the same `unredeemedShares` multiple
> times.

This was fixed by doing a max redeem before `initiateWithdraw`

PR #60, opened 2021-08-04
add steth vault basics

chuddster:


contracts/vaults/STETHVault/base/RibbonVault.sol
```diff
@@ -0,0 +1,633 @@
...
+
+    /**
+     * @notice Returns the vault's total balance, including the amounts locked into a short position
+     * @return total balance of the vault, including the amounts locked in third party protocols
+     */
+    function totalBalance() public view returns (uint256) {
+        return
+            uint256(vaultState.lockedAmount)
+                .add(IERC20(vaultParams.asset).balanceOf(address(this)))
+                .add(VaultLifecycleSTETH.dswmul(0, 1));
```
chuddster, line 600: do some steth -> eth conversion stuff

contracts/vaults/STETHVault/RibbonThetaSTETHVault.sol
```diff
@@ -0,0 +1,336 @@
...
+     * @notice Rolls the vault's funds into a new short position.
+     */
+    function rollToNextOption() external nonReentrant {
+        (address newOption, uint256 queuedWithdrawAmount) = _rollToNextOption();
+
+        // Locked balance denominated in `collateralToken`
+
+        uint256 lockedBalance =
+            collateralToken.balanceOf(address(this)).sub(
+                VaultLifecycleSTETH.dswdiv(queuedWithdrawAmount, 0)
```
chuddster, line 265: do some steth -> eth conversion rate stuff

contracts/interfaces/ISTETH.sol
```diff
@@ -0,0 +1,49 @@
...
+        external
+        returns (uint256);
+
+    function withdraw(
+        uint256 _maxShares,
+        address _recipient,
+        uint256 _maxLoss
+    ) external returns (uint256);
+
+    function withdraw(uint256 _maxShares) external returns (uint256);
```
chuddster, line 17: get the right methods here.

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -0,0 +1,794 @@
...
+            closeParams.delay
+        );
+
+        premium = dswmul(
+            GnosisAuction.getOTokenPremium(
+                otokenAddress,
+                optionsPremiumPricer,
+                premiumDiscount
+            ),
+            ISTETH(collateralAsset).pricePerShare().mul(
```
chuddster, line 100: do steth -> eth stuff here

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -0,0 +1,794 @@
...
+        address weth,
+        address asset,
+        address collateralToken,
+        address recipient,
+        uint256 amount
+    ) external returns (uint256 withdrawAmount) {
+        uint256 pricePerYearnShare = ISTETH(collateralToken).pricePerShare();
+        withdrawAmount = dswdiv(
+            amount,
+            pricePerYearnShare.mul(decimalShift(collateralToken))
```
chuddster, line 540: do steth -> eth stuff here

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -0,0 +1,794 @@
...
+        address asset,
+        address collateralToken
+    ) external {
+        uint256 assetBalance = IERC20(asset).balanceOf(address(this));
+        ISTETH collateral = ISTETH(collateralToken);
+
+        uint256 amountToUnwrap =
+            dswdiv(
+                dsmax(assetBalance, amount).sub(assetBalance),
+                collateral.pricePerShare().mul(decimalShift(collateralToken))
```
chuddster, line 631: do steth -> eth stuff here

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -0,0 +1,794 @@
...
+            );
+
+        if (amountToUnwrap > 0) {
+            collateral.withdraw(amountToUnwrap);
```
chuddster, line 635: .withdraw() -> steth equivalent

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -0,0 +1,794 @@
...
+
+        if (amountToWrap > 0) {
+            IERC20(asset).safeApprove(collateralToken, amountToWrap);
+
+            // there is a slight imprecision with regards to calculating back from yearn token -> underlying
+            // that stems from miscoordination between ytoken .deposit() amount wrapped and pricePerShare
+            // at that point in time.
+            // ex: if I have 1 eth, deposit 1 eth into yearn vault and calculate value of yearn token balance
+            // denominated in eth (via balance(yearn token) * pricePerShare) we will get 1 eth - 1 wei.
+            ISTETH(collateralToken).deposit(amountToWrap, address(this));
```
chuddster, line 655: deposit -> steth equivalent

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -0,0 +1,794 @@
...
+            return;
+        }
+        IERC20(asset).safeTransfer(recipient, amount);
+    }
+
+    /**
+     * @notice Returns the decimal shift between 18 decimals and asset tokens
+     * @param collateralToken is the address of the collateral token
+     */
+    function decimalShift(address collateralToken)
```
chuddster, line 721: do we need decimal shift

contracts/interfaces/ISTETH.sol
```diff
@@ -0,0 +1,49 @@
...
+    function decimals() external view returns (uint256);
+}
+
+interface IYearnRegistry {
+    function latestVault(address token) external returns (address);
+}
+
+interface IYearnPricer {
+    function setExpiryPriceInOracle(uint256 _expiryTimestamp) external;
+}
```
kenchangh, line 49: Forgot to remove these

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -0,0 +1,794 @@
...
+            otoken.underlyingAsset() == vaultParams.underlying,
+            "Wrong underlyingAsset"
+        );
+        require(
+            otoken.collateralAsset() == collateralAsset,
+            "Wrong collateralAsset"
+        );
+
+        // we just assume all options use USDC as the strike
+        require(otoken.strikeAsset() == USDC, "strikeAsset != USDC");
```
kenchangh, line 129: We removed these checks entirely for VaultLifecycle.sol, feels like these checks are redundant because we're already passing them in as arguments for `factory.createOtoken`

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -0,0 +1,794 @@
...
+            (controller.getAccountVaultCounter(address(this))).add(1);
+
+        IOtoken oToken = IOtoken(oTokenAddress);
+        address collateralAsset = oToken.collateralAsset();
+
+        uint256 collateralDecimals =
+            uint256(IERC20Detailed(collateralAsset).decimals());
+        uint256 mintAmount;
+
+        if (oToken.isPut()) {
```
kenchangh, line 207: Does it make sense to remove all conditions checking `isPut`? stETH can only be used for call options.

Cons of doing so: The code across VaultLifecycles are not consistent

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -0,0 +1,794 @@
...
+     * @notice Returns the decimal shift between 18 decimals and asset tokens
+     * @param collateralToken is the address of the collateral token
+     */
+    function decimalShift(address collateralToken)
+        public
+        view
+        returns (uint256)
+    {
+        return
+            10**(uint256(18).sub(IERC20Detailed(collateralToken).decimals()));
```
kenchangh, line 727: Given the fact that VaultLifecycleSTETH is only *limited* to stETH, we don't have to call `decimals()` at all and just 18

We can even remove the decimalShift function entirely because `10**(18-18) = 1`

contracts/vaults/STETHVault/base/RibbonVault.sol
chuddster: 0, 1 is placeholder

contracts/libraries/VaultLifecycleSTETH.sol
chuddster: I initially thought they are redundant, but actually opyn can change code for token blueprint to create a different token. less attack surface this way

contracts/libraries/VaultLifecycleSTETH.sol
chuddster: we can remove it

contracts/libraries/VaultLifecycleSTETH.sol
kenchangh: Okay, let's do a follow-up PR to add those back into VaultLifecycle.sol

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -0,0 +1,747 @@
...
+            IWSTETH(collateralToken).getWstETHByStETH(
+                dsmax(assetBalance, amount).sub(assetBalance)
+            );
+
+        if (amountToUnwrap > 0) {
+            IWSTETH wsteth = IWSTETH(collateralToken);
+            // Unrap to stETH
+            wsteth.unwrap(wsteth.balanceOf(address(this)));
+
+            // CRV SWAP HERE from steth -> eth
```
chuddster, line 607: need to do crv swap from steth -> eth. 

withdrawals from lido directly not allowed for now. will need to upgrade when it is available

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -0,0 +1,747 @@
...
+
+    /**
+     * @notice Wraps the necessary amount of the base token to the yield-bearing yearn token
+     * @param collateralToken is the address of the collateral token
+     */
+    function wrapToYieldToken(address collateralToken) external {
+        ISTETH steth = ISTETH(IWSTETH(collateralToken).stETH());
+        if (address(this).balance > 0) {
+            // Send eth to Lido, recieve steth
+            steth.submit{value: address(this).balance}(address(this));
```
chuddster, line 619: passing in address(this) which is the referral address. will be used for rewards program

contracts/interfaces/ISTETH.sol
```diff
@@ -0,0 +1,87 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.7.2;
+
+interface IWSTETH {
+    function getStETHByWstETH(uint256 _amount) external view returns (uint256);
```
chuddster, line 5: please use for reference for getStETHByWstETH, getWstETHByStETH, etc.
wsteth: https://etherscan.io/address/0x7f39c581f595b53c5cb19bd0b3f8da6c935e2ca0#code
steth: https://etherscan.io/address/0x20dc62d5904633cc6a5e34bec87a048e80c92e97#code

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -0,0 +1,747 @@
...
+
+        uint256 amountToUnwrap =
+            IWSTETH(collateralToken).getWstETHByStETH(
+                dsmax(assetBalance, amount).sub(assetBalance)
+            );
+
+        if (amountToUnwrap > 0) {
+            IWSTETH wsteth = IWSTETH(collateralToken);
+            // Unrap to stETH
+            wsteth.unwrap(wsteth.balanceOf(address(this)));
```
chuddster: you mentioned high gas bc we are swapping. we could potentially swap more than necessary (like 1.25x) so that we have more eth on hand. 

do we prioritize maximal efficiency or gas fees for user withdrawals.

contracts/vaults/STETHVault/base/RibbonVault.sol
```diff
@@ -0,0 +1,632 @@
...
+        require(msg.value > 0, "!value");
+
+        _depositFor(msg.value, msg.sender);
+    }
+
+    /**
+     * @notice Deposits the `collateralAsset` into the contract and mint vault shares.
+     * @param amount is the amount of `collateralAsset` to deposit
+     */
+    function deposit(uint256 amount) external nonReentrant {
```
chuddster, line 203: this now does steth deposit

contracts/libraries/VaultLifecycleSTETH.sol
kenchangh: I'm fine with how we're doing it now, as long we don't go over 200k gas it should be fine

contracts/vaults/STETHVault/base/RibbonVault.sol
kenchangh: Rename to `depositStETH` for clarity?

kenchangh: or `depositYieldToken`

contracts/vaults/STETHVault/base/RibbonVault.sol
```diff
@@ -0,0 +1,632 @@
...
+
+        // If we have an unprocessed pending deposit from the previous rounds, we have to process it.
+        uint128 unredeemedShares =
+            depositReceipt.getSharesFromReceipt(
+                currentRound,
+                roundPricePerShare[depositReceipt.round],
+                vaultParams.decimals
+            );
+
+        uint256 depositAmount = uint104(amount);
```
kenchangh, line 242: Don't have to cast to uint104 here

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -0,0 +1,730 @@
...
+            ICRV crv = ICRV(crvPool);
+
+            // CRV SWAP HERE from steth -> eth
+            // 0 = ETH, 1 = STETH
+            uint256 minETHOut =
+                crv.get_dy(1, 0, amountToUnwrap).mul(
+                    uint256(256).sub(crvPoolSlippage).div(100)
+                );
+
+            crv.exchange(1, 0, amountToUnwrap, minETHOut);
```
kenchangh, line 602: This approach makes it vulnerable to sandwich attacks. To fix this, we need to pass in the `minETHOut` as a parameter for withdrawals

chuddster: can you elaborate? can't a bot still frontrun this tx even if we pass in some arbitrary minETHOut calculated off-chain?

contracts/vaults/STETHVault/base/RibbonVault.sol
```diff
@@ -0,0 +1,650 @@
...
+    using SafeMath for uint256;
+    using ShareMath for Vault.DepositReceipt;
+
+    /************************************************
+     *  IMMUTABLES & CONSTANTS
+     ***********************************************/
+
+    address public immutable USDC;
+    address public immutable WETH;
+    address public immutable LDO = 0x5A98FcBEA516Cf06857215779Fd812CA3beF1B32;
```
kargakis, line 28: Any reason this is not provided via the constructor like the rest of the tokens?

contracts/vaults/STETHVault/base/RibbonVault.sol
```diff
@@ -0,0 +1,661 @@
...
+     * @param isETH is whether this is a depositETH call
+     */
+    function _depositFor(
+        uint256 amount,
+        address creditor,
+        bool isETH
+    ) private {
+        uint256 currentRound = vaultState.round;
+        uint256 totalWithDepositedAmount =
+            isETH ? totalBalance() : totalBalance().add(amount);
```
chuddster, line 240: added an isETH param because totalBalance() for eth deposits will have already included the deposit (msg.value not transerFrom), so we double count amount.

alternative is remove isETH, make it totalWithDepositedAmount = totalBalance(), and do yieldToken.safeTransferFrom before _depositFor in depositYieldToken

contracts/vaults/STETHVault/base/RibbonVault.sol
```diff
@@ -0,0 +1,682 @@
...
+        transferOwnership(_owner);
+
+        keeper = _keeper;
+
+        feeRecipient = _feeRecipient;
+        performanceFee = _performanceFee;
+        managementFee = _managementFee.mul(10**6).div(WEEKS_PER_YEAR);
+        vaultParams = _vaultParams;
+
+        collateralToken = IWSTETH(0x7f39C581F595B53c5cb19bD0b3f8dA6c935E2Ca0);
```
chuddster, line 150: hardcoding tad bit unprofessional. but best way without getting stack too deep error in initialize() and not adding wsteth as variable in contract (redundant to collateralToken)

contracts/vaults/STETHVault/RibbonThetaSTETHVault.sol
```diff
@@ -0,0 +1,353 @@
...
+
+        auctionDuration = newAuctionDuration;
+    }
+
+    /**
+     * @notice Sets the new strike selection or options premium pricer contract
+     * @param newContract is the address of the new strike selection or options premium pricer contract
+     * @param isStrikeSelection is whether we are setting the strike selection contract
+     */
+    function setStrikeSelectionOrPricer(
```
chuddster, line 165: condensed setStrikeSelection / setOptionsPremiumPricer to one method to fit contract size constraint

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -0,0 +1,743 @@
...
+        }
+
+        if (address(this).balance > 0) {
+            // Send eth to Lido, recieve steth
+            ISTETH(steth).submit{value: address(this).balance}(address(this));
+            uint256 stethBalance = stethToken.balanceOf(address(this));
+            // approve wrap
+            stethToken.doubleApprove(collateralToken, stethBalance.add(1));
+            // Wrap to wstETH - need to add 1 to steth balance as it is innacurate
+            collateral.wrap(stethBalance.add(1));
```
chuddster, line 628: for some reason stethBalance is off by 1 wei. so adding one to wrap. (I checked steth balance after wrap and it was 1)

test/RibbonThetaSTETHVault.ts
```diff
@@ -0,0 +1,2907 @@
...
+      });
+    });
+
+    describe("#withdrawInstantly", () => {
+      let minETHOut: BigNumberish;
+
+      time.revertToSnapshotAfterEach(async function () {
+        const crv = await getContractAt("ICRV", STETH_ETH_CRV_POOL);
+        minETHOut = (await crv.get_dy(1, 0, depositAmount))
+          .mul(crvSlippage.add(100))
```
chuddster, line 2203: for some reason transaction reverts on crv.get_dy(1, 0, depositAmount) do you have any idea why? I checked on the actual contract https://etherscan.io/address/0xDC24316b9AE028F1497c275EB9192a3Ea0f67022#readContract and everything works...

test/RibbonThetaSTETHVault.ts
```diff
@@ -0,0 +1,2907 @@
...
+          method: "hardhat_impersonateAccount",
+          params: [LDO_HOLDER],
+        });
+        const ldoHolder = await ethers.provider.getSigner(LDO_HOLDER);
+
+        const ldo = await ethers.getContractAt("IERC20", LDO_ADDRESS);
+
+        let ldoDepositAmount = BigNumber.from(100);
+
+        await ldo.connect(ldoHolder).transfer(vault.address, ldoDepositAmount);
```
chuddster, line 2692: transaction reverts on .transfer() of ldo tokens. do you see offhand anything I am doing wrong? EOA has ldo and has eth

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -0,0 +1,747 @@
...
+     * @notice Withdraws yvWETH from vault
+     * @param collateralToken is the address of the collateral token
+     * @param recipient is the recipient
+     * @param withdrawAmount is the withdraw amount in terms of yearn tokens
+     */
+    function withdrawYieldToken(
+        address collateralToken,
+        address recipient,
+        uint256 withdrawAmount
+    ) internal returns (uint256 yieldTokenBalance) {
```
kenchangh, line 538: Needs to be external so the ThetaVault contract does not end up importing the code from the library

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -0,0 +1,747 @@
...
+     * @param yieldTokenBalance is the collateral token (yvWETH) balance of the vault
+     */
+    function withdrawBaseToken(
+        address weth,
+        address asset,
+        address collateralToken,
+        address recipient,
+        uint256 withdrawAmount,
+        uint256 yieldTokenBalance
+    ) internal {
```
kenchangh, line 561: https://github.com/ribbon-finance/ribbon-v2/pull/60/files/683f5165ef1e8ab1f725b8375451429b5a8fee03#r692417777

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -0,0 +1,747 @@
...
+        uint256 singleShare
+    ) internal pure returns (uint256 newPricePerShare) {
+        newPricePerShare = currentSupply > 0
+            ? singleShare.mul(roundStartBalance).div(currentSupply)
+            : singleShare;
+    }
+
+    /***
+     * DSMath Copy paste
+     */
```
kenchangh, line 714: Seeing as we copy paste this everywhere, let's just bite the bullet and make this a separate library file instead

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -0,0 +1,747 @@
...
+        // deposits and withdrawals, is positive. If it is negative, last week's
+        // option expired ITM past breakeven, and the vault took a loss so we
+        // do not collect performance fee for last week
+        if (currentLockedBalance.sub(totalPending) > prevLockedAmount) {
+            performanceFee = currentLockedBalance
+                .sub(totalPending)
+                .sub(prevLockedAmount)
+                .mul(performanceFeePercent)
+                .div(100 * 10**6);
+            managementFee = currentLockedBalance.mul(managementFeePercent).div(
```
kenchangh, line 661: Needs to do `.sub(totalPending)` as per OZ's feedback

contracts/libraries/VaultLifecycleSTETH.sol
kenchangh: `address(this).balance` actually consumes a lot of gas (400). We need to cache it in a variable

Screenshot 2021-08-19 at 9 32 32 PM

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -0,0 +1,747 @@
...
+     * @notice Wraps the necessary amount of the base token to the yield-bearing yearn token
+     * @param collateralToken is the address of the collateral token
+     */
+    function wrapToYieldToken(address collateralToken) external {
+        ISTETH steth = ISTETH(IWSTETH(collateralToken).stETH());
+        if (address(this).balance > 0) {
+            // Send eth to Lido, recieve steth
+            steth.submit{value: address(this).balance}(address(this));
+            // Wrap to wstETH
+            IWSTETH(collateralToken).wrap(steth.balanceOf(address(this)));
```
kenchangh: Optimization: We can use the value returned by `submit` to get the number of shares returned. This is to avoid calling `steth.balanceOf`, which consumes 2100 gas.

```sol
uint shares = steth.submit{value: ethBalance}(address(this));
IWSTETH(collateralToken).wrap(shares); // double check this
```

https://etherscan.io/address/0x20dc62d5904633cc6a5e34bec87a048e80c92e97#code

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -0,0 +1,747 @@
...
+            require(success, "!success");
+            return;
+        }
+        IERC20(asset).safeTransfer(recipient, amount);
+    }
+
+    /**
+     * @notice Gets the next options expiry timestamp
+     */
+    function getNextFriday(uint256 currentExpiry)
```
kenchangh, line 682: This function needs to be copy paste the updated version from VaultLifecycle.sol

contracts/libraries/VaultLifecycleSTETH.sol
chuddster: actually needs to be internal because withdrawYieldToken / withdrawBaseToken is used within vault lifecycle methods.

get "withdrawYieldToken" is not (or not yet) visible at this point. otherwise

contracts/libraries/VaultLifecycleSTETH.sol
chuddster: I want to make all of the changes based on OZ feedback in a completely different pr. otherwise its a mess to coordinate what has and what has not been addressed. 

i remember :)

contracts/vaults/STETHVault/base/RibbonVault.sol
kenchangh: How would we be able to test this on Kovan if we hardcoded it?

contracts/libraries/VaultLifecycleSTETH.sol
chuddster: yeah, want to do this all in separate pr for leftover chainsafe changes to be addressed :)

test/RibbonThetaSTETHVault.ts
kenchangh: Nothing stands out to me why it reverts..

kenchangh: Does the pool exist when the block is forked

test/RibbonThetaSTETHVault.ts
kenchangh: It could be related to the LDO vesting. Check out the [doTransfer function in LDO](https://etherscan.io/address/0x5a98fcbea516cf06857215779fd812ca3bef1b32#code), which checks the [Controller](https://etherscan.io/address/0xf73a1260d222f447210581ddf212d915c09a3249) if an account can spend their LDOs or not.

Screenshot 2021-08-19 at 10 22 27 PM

Try getting a smaller LDO retail holder

test/RibbonThetaSTETHVault.ts
chuddster: 13056027 is 3 days ago...

contracts/libraries/VaultLifecycleSTETH.sol
chuddster: actually, cant do this because what if somebody deposits steth directly.

contracts/vaults/STETHVault/base/RibbonVault.sol
```diff
@@ -0,0 +1,685 @@
...
+
+        uint256 withdrawAmount =
+            ShareMath.sharesToUnderlying(
+                withdrawalShares,
+                roundPricePerShare[uint16(withdrawalRound)],
+                vaultParams.decimals
+            );
+
+        // Unwrap may incur curve pool slippage
+        uint256 amountETHOut =
```
chuddster, line 377: the amount we transfer takes into account curve pool slippage - so they will get less. 

same for withdrawInstantly()

contracts/vaults/STETHVault/base/RibbonVault.sol
```diff
@@ -0,0 +1,685 @@
...
+        );
+
+        _transfer(msg.sender, address(this), shares);
+    }
+
+    /**
+     * @notice Completes a scheduled withdrawal from a past round. Uses finalized pps for the round
+     * @param minETHOut is the min amount of `asset` to recieve for the swapped amount of steth in crv pool
+     */
+    function completeWithdraw(uint256 minETHOut) external nonReentrant {
```
chuddster, line 352: just a note for when jhh does frontend - minETHOut will need to do extensive math:

1. [shareToUnderlying](https://github.com/ribbon-finance/ribbon-v2/pull/60/files#diff-5020f0dced6f2a6457e6f7cc1592eb79cf21080973d4fab3500f747feab91deaR369)
2. [getWstETHByStETH](https://github.com/ribbon-finance/ribbon-v2/pull/60/files#diff-2bb6f455514e2f789d18593144b618eac195c36fec6d89dd1200da256764c48dR589)
3. get_dy(1, 0, ABOVE_CALCULATION) * slippage / 100

contracts/vaults/STETHVault/base/RibbonVault.sol
chuddster: we can just create a kovan copy. its ugly but given constraints its best.

test/RibbonThetaSTETHVault.ts
chuddster: this was it!

contracts/vaults/STETHVault/base/RibbonVault.sol
```diff
@@ -0,0 +1,685 @@
...
+    /**
+     * @notice Deposits the `collateralAsset` into the contract and mint vault shares.
+     * @param amount is the amount of `collateralAsset` to deposit
+     */
+    function depositYieldToken(uint256 amount) external nonReentrant {
+        require(amount > 0, "!amount");
+
+        _depositFor(
+            // off by one
+            collateralToken.getStETHByWstETH(amount).sub(1),
```
chuddster, line 237: off by 1

contracts/vaults/STETHVault/RibbonThetaSTETHVault.sol
```diff
@@ -0,0 +1,356 @@
...
+        );
+
+        // Subtraction underflow checks already ensure it is smaller than uint104
+        depositReceipt.amount = uint104(uint256(receiptAmount).sub(amount));
+
+        emit InstantWithdraw(msg.sender, amount, currentRound);
+
+        // Unwrap may incur curve pool slippage
+        uint256 amountETHOut =
+            VaultLifecycleSTETH.unwrapYieldToken(
```
kenchangh, line 210: Why would you need to do `unwrapYieldToken` when it's `withdrawInstantly`? Shouldn't the token still be WETH since it's not rolled yet?

chuddster: firstly, it would be eth (no weth deposits). 

secondly, you are assuming I am depositing eth. what if I am depositing steth? we don't allow for withdrawYieldToken() so we have to do this check. _most likely_ in the future there will be enough other eth deposits in the same round such that we wont actually have to unwrap. so unwrapYieldToken will just exit out after a few lines. but in the off chance we need to unwrap, we need this.

contracts/vaults/STETHVault/RibbonThetaSTETHVault.sol
```diff
@@ -0,0 +1,356 @@
...
+            );
+
+        if (unlockedAssedAmount > 0) {
+            vaultState.lockedAmount = uint104(
+                uint256(vaultState.lockedAmount).sub(unlockedAssedAmount)
+            );
+        }
+
+        // Wrap entire `asset` balance to `collateralToken` balance
+        VaultLifecycleSTETH.wrapToYieldToken(WETH, address(collateralToken));
```
kenchangh, line 340: Why do we have to `wrapToYieldToken` if we are receiving wstETH after burning?

chuddster: for the same reason we do it in yearn. 

burnRemainingOTokens() is one of the functions we 100% know is called after the auction, hence after we received premiums for this week. so if we do end up having to burn otokens, why not maximize efficiency and wrap the premium to yield bearing token right then and there? otherwise it lays idle for a week.

kenchangh: Oh I see - I did not consider the premiums that will be in WETH.

test/RibbonThetaSTETHVault.ts
```diff
@@ -0,0 +1,2943 @@
...
+
+        // 100% of the vault's balance is allocated to short
+        assert.bnEqual(
+          await assetContract.balanceOf(vault.address),
+          BigNumber.from(0)
+        );
+
+        await expect(
+          vault.connect(ownerSigner).commitAndClose()
+        ).to.be.revertedWith("C31");
```
kenchangh, line 1603: What is C31? Is it an internal Opyn error?

contracts/vaults/STETHVault/base/RibbonVault.sol
```diff
@@ -0,0 +1,685 @@
...
+            );
+
+        // Unwrap may incur curve pool slippage
+        uint256 amountETHOut =
+            VaultLifecycleSTETH.unwrapYieldToken(
+                withdrawAmount,
+                address(collateralToken),
+                STETH_ETH_CRV_POOL,
+                minETHOut
+            );
```
kenchangh, line 383: `require(amountETHOut > 0)`

contracts/vaults/STETHVault/base/RibbonVault.sol
```diff
@@ -0,0 +1,685 @@
...
+
+        VaultLifecycleSTETH.transferAsset(msg.sender, amountETHOut);
+
+        require(withdrawAmount > 0, "!withdrawAmount");
```
kenchangh, line 387: Move this require check above the `transferAsset`

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -0,0 +1,758 @@
...
+                dsmax(assetBalance, amount).sub(assetBalance)
+            );
+
+        if (amountToUnwrap > 0) {
+            IWSTETH wsteth = IWSTETH(collateralToken);
+            // Unrap to stETH
+            wsteth.unwrap(amountToUnwrap);
+
+            // approve steth exchange
+            IERC20(wsteth.stETH()).doubleApprove(crvPool, amountToUnwrap);
```
kenchangh, line 599: Can avoid using `doubleApprove` completely because we are only using stETH, which is not a problematic ERC20

test/RibbonThetaSTETHVault.ts
chuddster: look at line 1223 with controller codes https://etherscan.io/address/0xcc2fd280a669a4453b30b5e72871ad3e2a2663d0/advanced#code

its shorthand for the errors we had before but thats what it reverted with

contracts/libraries/VaultLifecycleSTETH.sol
chuddster: fixed

kenchangh (dismissed):

kenchangh (approved):

kenchangh (dismissed):

kenchangh (approved):

PR #61, opened 2021-08-05
Yearn deposit yield token

chuddster:


test/RibbonThetaYearnVault.ts
```diff
@@ -933,92 +933,169 @@ function behavesLikeRibbonOptionsVault(params: {
       });
     });
```
chuddster, line 935: need to finish

test/RibbonThetaYearnVault.ts
```diff
@@ -83,8 +83,8 @@ describe("RibbonThetaYearnVault", () => {
     tokenDecimals: 18,
     isPut: false,
     gasLimits: {
-      depositWorstCase: 140390,
-      depositBestCase: 122555,
+      depositWorstCase: 154539,
+      depositBestCase: 133598,
```
kenchangh, line 87: Question out of scope of this PR, why is the gas use for deposits so much [worse than the vanilla vaults](https://github.com/ribbon-finance/ribbon-v2/blob/master/test/RibbonThetaVault.ts#L79-L80)?

kenchangh (approved):
LGTM

test/RibbonThetaYearnVault.ts
kargakis: :(

PR #64, opened 2021-08-10
Kovan e2e test scripts

kenchangh:
- Can safely ignore the `deployments/*.json` files, all is summarized in `constants/deployments.json`.
- Most importantly, introduced a E2E test script for us to run manually to sanity check all the vault functions


package.json
```diff
@@ -12,7 +12,7 @@
     "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-prettier": "^3.4.0",
     "ethereum-waffle": "^3.0.0",
-    "ethers": "^5.0.0",
+    "ethers": "5.0.0",
```
kenchangh, line 15: Had to downgrade Ethers because the EIP1559 support was bugging out transactions in the E2E tests (Kovan does not support the new tx types yet)

chuddster (approved):

PR #72, opened 2021-08-22
make chainsafe fixes

chuddster:


contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -0,0 +1,760 @@
...
+    /**
+     * @notice Gets the next options expiry timestamp
+     * @param currentExpiry is the expiry timestamp of the current option
+     * Reference: https://codereview.stackexchange.com/a/33532
+     * Examples:
+     * getNextFriday(week 1 thursday) -> week 1 friday
+     * getNextFriday(week 1 friday) -> week 2 friday
+     * getNextFriday(week 1 saturday) -> week 2 friday
+     */
+    function getNextFriday(uint256 currentExpiry)
```
chuddster, line 702: add getNextFriday

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -0,0 +1,757 @@
...
+        IStrikeSelection selection = IStrikeSelection(strikeSelection);
+
+        (strikePrice, delta) = closeParams.lastStrikeOverride ==
+            vaultState.round
+            ? (closeParams.overriddenStrikePrice, selection.delta())
+            : selection.getStrikePrice(expiry, false);
+
+        require(strikePrice != 0, "!strikePrice");
+
+        otokenAddress = getOrDeployOtoken(
```
chuddster, line 77: there are places like this that are different from vaultlifecycle - verifyOtoken is outside getOrDeployOToken. 

vaultlifecycle - [verifyOtoken is inside getOrDeployOToken ](https://github.com/ribbon-finance/ribbon-v2/blob/512b9573c8d71f717a460b6006236fc3ec29166f/contracts/libraries/VaultLifecycle.sol#L75)

this is a harmful difference. tests work no need to mess w it now

PR #81, opened 2021-08-27
[M02] Rounds with no locked asset can lead to fee miscalculations

chuddster:


contracts/vaults/YearnVaults/RibbonThetaYearnVault.sol
```diff
@@ -0,0 +1,380 @@
...
+        );
+    }
+
+    /**
+     * @notice Optionality to set strike price manually
+     * @param strikePrice is the strike price of the new oTokens (decimals = 8)
+     */
+    function setStrikePrice(uint128 strikePrice)
+        external
+        onlyOwner
```
chuddster, line 373: disregard everything yearn and steth

PR #85, opened 2021-08-30
[H01] Corruptible storage upgradeability pattern

kenchangh:
- OptionsVaultStorage now part of RibbonVault.sol, and is not upgradeable
- Refactored RibbonThetaVaultStorage and RibbonDeltaVaultStorage into different files for readability
- Added gap as safety precaution


contracts/vaults/base/RibbonVault.sol
```diff
@@ -18,12 +26,52 @@ import {
...
+    ReentrancyGuardUpgradeable,
+    OwnableUpgradeable,
+    ERC20Upgradeable
+{
     using SafeERC20 for IERC20;
     using SafeMath for uint256;
     using ShareMath for Vault.DepositReceipt;
 
+    /************************************************
+     *  NON UPGRADEABLE STORAGE
```
chuddster, line 41: perhaps we still move the storage variables to a different contract for readability?

chuddster (approved):

contracts/vaults/base/RibbonVault.sol
kenchangh: There isn't really a point of moving it into a separate contract because we don't intend to upgrade it. The inheritance chain also makes it hard to follow

PR #91, opened 2021-09-01
[L04] Duplicated price per share calculations

kenchangh:


contracts/vaults/base/RibbonVault.sol
```diff
@@ -567,13 +567,21 @@ contract RibbonVault is OptionsVaultStorage {
...
+                totalBalance(),
+                vaultState.totalPending,
+                _decimals
+            );
+
+        return
+            ShareMath.sharesToUnderlying(
+                shares(account),
+                pps,
+                uint8(_decimals)
```
kenchangh, line 583: [This PR](https://github.com/ribbon-finance/ribbon-v2/pull/88) needs to be merged before L04. We're casting to uint8 because we're changing the parameter to uint256 in a separate PR

Will remove the cast after that's merged

PR #93, opened 2021-09-01
[L07] Magic numbers are used

kenchangh:


contracts/vaults/base/RibbonVault.sol
```diff
@@ -447,7 +464,7 @@ contract RibbonVault is OptionsVaultStorage {
      * @param numRounds is the number of rounds to initialize in the map
      */
     function initRounds(uint256 numRounds) external nonReentrant {
-        require(numRounds < 52, "numRounds >= 52");
+        require(numRounds > 0, "!numRounds");
```
kenchangh, line 467: Removing 52 entirely

PR #99, opened 2021-09-02
[N01] Broken links in inline documentation

chuddster:


contracts/libraries/VaultLifecycle.sol
```diff
@@ -151,7 +151,7 @@ library VaultLifecycle {
         );
     }
```
chuddster, line 153: [N02] Implicit access control has drawbacks - fixed by using onlyKeeper()

PR #114, opened 2021-09-03
burn only after settle, commitandclose later

chuddster:


scripts/masterCron.ts
```diff
@@ -562,7 +568,7 @@ async function run() {
 
   var commitAndCloseJob = new CronJob(
     // 0 0 10 * * 5 = 10am UTC on Fridays.
-    `0 0 ${COMMIT_START} * * 5`,
+    `0 ${NETWORK_CONGESTION_BUFFER * 3} ${COMMIT_START} * * 5`,
```
kenchangh, line 571: Can we just specify the exact minutes it will execute? Not sure multiplying the `NETWORK_CONGESTION_BUFFER` helps with readability

scripts/masterCron.ts
```diff
@@ -537,7 +542,7 @@ async function run() {
   client.login(process.env.DISCORD_TOKEN);
 
   //Atlantic/Reykjavik corresponds to UTC
-
+  const OPYN_PRICE_FINALIZATION_BUFFER = 15; // 5 minutes
```
kenchangh, line 545: comments say 5 minutes but the variable is 15?

kenchangh (approved):

PR #119, opened 2021-09-08
update token list to github token list repo

chuddster:
https://github.com/steveukx/git-js


scripts/masterCron.ts
```diff
@@ -86,6 +88,42 @@ const getNextFriday = (currentExpiry: number) => {
...
+    maxConcurrentProcesses: 6,
+  };
+
+  // when setting all options in a single object
+  const git: SimpleGit = simpleGit(options);
+
+  fs.writeFile(`../ribbon-token-list/${fileName}`, tokenlist);
+
+  await git
+    .cwd("../ribbon-token-list")
```
chuddster, line 119: setting current working directory (cwd) to other repo 
https://github.com/steveukx/git-js#changing-the-working-directory-examples

scripts/masterCron.ts
```diff
@@ -86,6 +88,42 @@ const getNextFriday = (currentExpiry: number) => {
...
+    .addConfig("user.name", "cron job")
+    .addConfig("user.email", "some@one.com")
```
kenchangh, line 121: Need to add actual credentials through .env

scripts/masterCron.ts
```diff
@@ -86,6 +88,42 @@ const getNextFriday = (currentExpiry: number) => {
...
+  const options: Partial<SimpleGitOptions> = {
+    baseDir: process.cwd(),
+    binary: "git",
+    maxConcurrentProcesses: 6,
+  };
+
+  // when setting all options in a single object
+  const git: SimpleGit = simpleGit(options);
+
+  fs.writeFile(`../ribbon-token-list/${fileName}`, tokenlist);
```
kenchangh, line 116: This path won't resolve correctly because `..` would be `./scripts`. Best to hardcode a path like `/home/ribbon-token-list/`

scripts/masterCron.ts
```diff
@@ -86,6 +88,42 @@ const getNextFriday = (currentExpiry: number) => {
...
+
+  await git
+    .cwd("../ribbon-token-list")
+    .addConfig("user.name", "cron job")
+    .addConfig("user.email", "some@one.com")
+    .add(fileName)
+    .commit("update tokenlist")
```
kenchangh, line 123: Add the date into the commit so we can look back on faulty commits

scripts/masterCron.ts
```diff
@@ -211,6 +249,47 @@ async function getAnnualizedVol(underlying: string, resolution: number) {
...
+async function updateTokenList(
+  fileName: string,
+  vaultArtifactAbi: any,
+  ierc20Abi: any,
+  provider: any,
+  network: string
+) {
+  let tokens = [];
+
+  let token = {
```
kenchangh, line 261: Use types to save us from bugs!

```typescript
interface Token {
 ...
}
```

scripts/masterCron.ts
```diff
@@ -211,6 +249,47 @@ async function getAnnualizedVol(underlying: string, resolution: number) {
...
+      ierc20Artifact.abi,
+      provider
+    );
+
+    token["address"] = oToken.address;
+    token["name"] = await oToken.name();
+    token["symbol"] = await oToken.symbol();
+    token["decimals"] = await oToken.decimals();
+
+    tokens += token;
```
kenchangh, line 287: This is better re-written by declaring the variable in the loop itself:

```typescript
const token: Token = {
 ...
}
tokens.push(token); // The '+' operator creates weird behaviors
```

scripts/masterCron.ts
```diff
@@ -211,6 +249,47 @@ async function getAnnualizedVol(underlying: string, resolution: number) {
...
 }
 
+async function updateTokenList(
+  fileName: string,
+  vaultArtifactAbi: any,
+  ierc20Abi: any,
+  provider: any,
+  network: string
+) {
+  let tokens = [];
```
kenchangh, line 259: Shouldn't we read the original list from the file and append to it? Instead of creating an empty array each time

scripts/masterCron.ts
```diff
@@ -211,6 +249,47 @@ async function getAnnualizedVol(underlying: string, resolution: number) {
...
+  };
+
+  for (let vaultName in deployments[network].vaults) {
+    const vault = new ethers.Contract(
+      deployments[network].vaults[vaultName].address,
+      vaultArtifactAbi,
+      provider
+    );
+
+    let oToken = new ethers.Contract(
```
kenchangh, line 276: Since we're appending to the list of tokens in the file, we need to verify that the otoken doesn't already exist in the list

https://github.com/ribbon-finance/ribbon-v2/pull/119/files#r706706095

scripts/masterCron.ts
chuddster: this what you mean or is there something else

https://github.com/steveukx/git-js#authentication

scripts/masterCron.ts
chuddster: can do ../../ribbon-token-list instead no?

scripts/masterCron.ts
chuddster: I think we should create an empty array each time. previous list does not matter since those otokens are irrelevant at this point. think it should just be the otokens relevant for this current work instead of a neverending list.

scripts/masterCron.ts
chuddster: see above

scripts/masterCron.ts
kenchangh: Yep. You can pass in the username and password to the function. But the credentials need to be stored in .env

scripts/masterCron.ts
kenchangh: Just arguing for maintainability because if we shift the script around it will break the path if we are using a relative path

scripts/masterCron.ts
kenchangh: Oh yeah, I didn't consider this. You're right.

kenchangh: My take on this is that we should have a never-ending list because users can browse through the past auctions and see the Ribbon symbol in them.

scripts/masterCron.ts
chuddster: agree this is better

scripts/masterCron.ts
chuddster: actually don't need to clone each time. can just clone once and set up the personal access token for the bot account for pushing. _should_ just be able to push in the script without any extra config right.

scripts/masterCron.ts
chuddster: fixed

scripts/masterCron.ts
chuddster: fixed

scripts/masterCron.ts
chuddster: fixed

scripts/masterCron.ts
```diff
@@ -113,14 +136,28 @@ async function pushTokenListToGit(tokenlist: string, fileName: string) {
...
-  fs.writeFile(`../ribbon-token-list/${fileName}`, tokenlist);
+  let newTokenSet = tokenSet;
+
+  let currentTokenSet = (
+    JSON.parse(fs.readFileSync(`../ribbon-token-list/${fileName}`)) as TokenSet
+  ).tokens;
+
+  // add new week's otokens to token list and remove duplicates
+  newTokenSet.tokens = Array.from(
+    new Set(currentTokenSet.concat(newTokenSet.tokens))
```
kenchangh, line 147: Just confirming, does `Set` filter out the duplicate objects? Normally only works when it's not an object

kenchangh (approved):

kenchangh (approved):

PR #124, opened 2021-09-09
fix redeem to not revert on shares = 0

chuddster:


test/RibbonThetaSTETHVault.ts
```diff
@@ -1985,7 +1985,29 @@ function behavesLikeRibbonOptionsVault(params: {
...
+        const { round, amount, unredeemedShares } = await vault.depositReceipts(
+          user
+        );
+
+        assert.equal(round, 1);
+        assert.bnEqual(amount, BigNumber.from(0));
+        assert.bnEqual(unredeemedShares, BigNumber.from(0));
+
+        await vault.maxRedeem();
+
```
kenchangh, line 2004: Can use this trick to double check that there were no transfers:

```
await expect(res).to.not.emit(vault, "Transfer");
```

kenchangh (approved):
LGTM

PR #132, opened 2021-09-17
Update cron token list update and burnOToken

chuddster:


scripts/masterCron.ts
```diff
@@ -353,11 +354,22 @@ async function settleAndBurn(
           .settleAuction(auctionID.toString(), {
             gasPrice: newGasPrice,
             gasLimit: gasLimits["settleAuction"],
-          });
+          })
+          .wait();
```
chuddster, line 358: we want to wait for transaction to be mined before checking otoken balance

scripts/masterCron.ts
```diff
@@ -558,12 +570,16 @@ async function strikeForecasting() {
...
 async function commitAndClose() {
   const vaultArtifact = await hre.artifacts.readArtifact("RibbonThetaVault");
-  const ierc20Artifact = await hre.artifacts.readArtifact(
-    "contracts/interfaces/IERC20Detailed.sol:IERC20Detailed"
-  );
 
   // 1. commitAndClose
   await runTX(vaultArtifact.abi, provider, signer, network, "commitAndClose");
+}
+
```
chuddster, line 577: decided to move token list update in rolltonextoption cron job. 

the reason is that in commitAndClose it will revert since I update token set without waiting for actual commitAndClose() to be mined (wait()). I could move it back and just wait() but then commitAndClose does not get called asynchronously so it is slower.

kenchangh (approved):

kenchangh (approved):

PR #134, opened 2021-09-24
Update script to lower gas price

kenchangh:


scripts/masterCron.ts
```diff
@@ -354,8 +354,9 @@ async function settleAndBurn(
           .settleAuction(auctionID.toString(), {
             gasPrice: newGasPrice,
             gasLimit: gasLimits["settleAuction"],
-          })
-          .wait();
+          });
+
+        await tx.wait();
```
kenchangh, line 359: Separating out wait ensures it throws with the correct message if settleAuction fails

chuddster (approved):

PR #148, opened 2021-10-08
Update discord messages

kenchangh:


scripts/masterCron.ts
```diff
@@ -598,13 +631,25 @@ async function rollToNextOption() {
...
-  let auctionCounters = await runTX(
-    vaultArtifact.abi,
-    provider,
-    signer,
-    network,
-    "rollToNextOption"
-  );
+  await runTX(vaultArtifact.abi, provider, signer, network, "rollToNextOption");
+
+  await sleep(5000);
```
chuddster, line 636: why are we sleeping?

scripts/masterCron.ts
```diff
@@ -577,9 +582,37 @@ async function strikeForecasting() {
...
+  const otokenArtifact = await hre.artifacts.readArtifact("IOtoken");
 
   // 1. commitAndClose
   await runTX(vaultArtifact.abi, provider, signer, network, "commitAndClose");
+
+  await sleep(5000);
+
+  let msg = `${auctionParticipantTag} Strike prices have been selected\n\n`;
+
+  for (let vaultName in deployments[network].vaults) {
```
chuddster, line 594: so is this to show the actual strike, etc. on top of the expected?

PR #151, opened 2021-10-18
Chainsafe Audit Changes

chuddster:
- withdraw from yearn vault correctly in upgradeYearnVault


contracts/libraries/Vault.sol
```diff
@@ -60,6 +60,8 @@ library Vault {
         uint128 totalPending;
         // Amount locked for scheduled withdrawals;
         uint128 queuedWithdrawShares;
+        // Amount locked for scheduled withdrawals last week;
+        uint128 lastQueuedWithdrawAmount;
```
kenchangh, line 64: We can't append `lastQueuedWithdrawAmount` into the `VaultState` because it will break the storage layout. So we need to append it into RibbonThetaVaultStorageV2 as a variable

```solidity
contract RibbonThetaVaultStorageV2 {
  uint256 lastQueuedWithdrawAmount;
}
```

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -322,7 +327,8 @@ library VaultLifecycleSTETH {
                 withdrawAmount.sub(yieldTokenBalance)
             );
 
-        transferAsset(payable(recipient), underlyingTokensToWithdraw);
+        IWETH(weth).deposit{value: underlyingTokensToWithdraw}();
+        IERC20(weth).safeTransfer(recipient, underlyingTokensToWithdraw);
```
kenchangh, line 331: What is the rationale behind this change?

contracts/vaults/BaseVaults/base/RibbonVault.sol
```diff
@@ -591,11 +591,22 @@ contract RibbonVault is
...
+            uint256(vaultState.lastQueuedWithdrawAmount);
+
         // Take management / performance fee from previous round and deduct
         lockedBalance = _lockedBalance.sub(
-            _collectVaultFees(_lockedBalance.add(queuedWithdrawAmount))
+            _collectVaultFees(
+                _lockedBalance.add(
+                    queuedWithdrawAmount > lastQueuedWithdrawAmount
+                        ? queuedWithdrawAmount.sub(lastQueuedWithdrawAmount)
+                        : 0
```
kenchangh, line 603: Probably makes sense to make this a variable like `withdrawAmountDiff` to help with readability

contracts/libraries/VaultLifecycleSTETH.sol
chuddster: if we use curve, the burner just does swapExactTokensForTokens. so figure it will be easier later down the line for RBN buyback to just have all ERC-20 tokens

contracts/libraries/Vault.sol
chuddster: addressed this

contracts/storage/RibbonThetaSTETHVaultStorage.sol
```diff
@@ -22,7 +22,7 @@ abstract contract RibbonThetaSTETHVaultStorageV1 {
 
 abstract contract RibbonThetaSTETHVaultStorageV2 {
     // Amount locked for scheduled withdrawals last week;
-    uint128 public lastQueuedWithdrawAmount;
+    uint256 public lastQueuedWithdrawAmount;
```
kenchangh, line 25: Changed it to uint256 to save gas

kenchangh (approved):

PR #155, opened 2021-11-11
Use different value for swap min

kenchangh:


contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -347,30 +347,63 @@ library VaultLifecycleSTETH {
...
+            // However it is fine because we check that the amountETHOut >= minETHOut at the end
+            // which makes sandwich attacks not possible
+            uint256 swappedAmount =
+                ICRV(crvPool).exchange(1, 0, stETHAmount, 1);
+
+            amountETHOut = amountETHOut.add(swappedAmount);
         }
 
+        // This revert does not account for the ETH that is already unwrapped
+        // Since minETHOut is derived from calling the Curve pool's getter,
```
chuddster, line 399: why do we want to add extra logic if .exchange() will revert for us already if stETHAmount < minETHOut if we pass that in .exchange()

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -347,30 +347,63 @@ library VaultLifecycleSTETH {
...
-            // Unrap to stETH
+            IERC20 steth = IERC20(wsteth.stETH());
+
+            uint256 startStethBalance = steth.balanceOf(address(this));
+
+            // Unwrap to stETH
             wsteth.unwrap(amountToUnwrap);
 
+            // Post-unwrap, the stETH balance will not completely match the stethNeeded
+            // due to precision issues.
```
chuddster, line 378: if we consistently know what the inconsistency will be we can just keep the way it is and not add more gas for user. but this is more dependable.

chuddster (approved):

PR #157, opened 2021-11-12
Chainsafe fixes

chuddster:
GnosisAuction.sol
65: Optimization, no need to poll balance of otokens again, reuse oTokenSellAmount variable instead.

VaultLifecycle.sol
_193: Minor, using newPricePerShare to calculate queuedWithdrawAmount is not excatly correct, as the shares were queued at a variety of prices._ - isnt this a significant issue if queuedWithdrawAmount is not exactly correct. by how much?
248: Note, the mintAmount value calculation could be simplified to (depositAmount*(10^Vault.OTOKEN_DECIMALS)*(10^Vault.OTOKEN_DECIMALS)) / (oToken.strikePrice() * (10^collateralDecimals))
**254: Minor, the collateralDecimals.sub(8) expression will revert on collateral with decimals < 8. The scaleBy calculation should be put inside of the if condition.**
_257: Note, the mintAmount value for collaterals with < 8 decimals should probably be scaledUp instead of down._ - do you mean in terms of proper coding or its wrong the way it is?
263: Note, the safeApproveNonCompliant call could be replaced with CustomSafeERC20.safeApprove.
**294: Note, 'deposited asset' comment should state 'option address'.**

VaultLifecycleSTETH.sol
_158: Minor, using newPricePerShare to calculate queuedWithdrawAmount is not excatly correct, as the shares were queued at a variety of prices._ - isnt this a significant issue if queuedWithdrawAmount is not exactly correct. by how much?
**205: Note, the double approve comment is outdated. Helper is handling it all now.**
**364: Major, specifying only the minETHOut is not enough, as the amountToUnwrap could be manipulated with frontrunning. Consider requiring amountETHOut >= minETHOut instead.**

VaultLifecycleYearn.sol
_164: Minor, using newPricePerShare to calculate queuedWithdrawAmount is not excatly correct, as the shares were queued at a variety of prices._ - isnt this a significant issue if queuedWithdrawAmount is not exactly correct. by how much?
_266: Major, invalid underlyingTokensToWithdraw calculation, should use ShareMath.sharesToAsset instead._ - can you elaborate more? thought we fixed it

BaseVaults\base\RibbonVault.sol
**263: Note, the setManagementFee function should emit ManagementFeeSet event.
288: Note, the setCap function should emit CapSet event.
524: Optimization, excessive assignment of depositReceipts[msg.sender].amount if condition is not met.
552: Optimization, the overflow check is not needed as the Solidity compiler 0.8+ performs it implicitly.
596: Major, adding the queuedWithdrawAmount to the locked balance for fee calculation is not always right. Consider a case last locked was 100 and queued 10, nothing changed in the current round, so current locked is 100 and queued is 10 again, the fee will be calculated from (100+10)-100 = 10. While there was no profit.**

BaseVaults\RibbonThetaVault.sol
**384: Note, the unlockedAssedAmount variable has a typo and should be named unlockedAssetAmount.**

STETHVault\base\RibbonVault.sol
**157: the 'LDO contract' comment should be replaced with 'wstETH contract'.
284: Note, the setManagementFee function should emit ManagementFeeSet event.
307: Note, the setCap function should emit CapSet event.
548: Optimization, excessive assignment of depositReceipts[msg.sender].amount if condition is not met.
576: Optimization, the overflow check is not needed as the Solidity compiler 0.8+ performs it implicitly.
622: Major, adding the queuedWithdrawAmount to the locked balance for fee calculation is not always right. Consider a case last locked was 100 and queued 10, nothing changed in the current round, so current locked is 100 and queued is 10 again, the fee will be calculated from (100+10)-100 = 10. While there was no profit.**

STETHVault\RibbonThetaSTETHVault.sol
**402: Note, the unlockedAssedAmount variable has a typo and should be named unlockedAssetAmount.**

YearnVaults\base\RibbonVault.sol
**281: Note, the setManagementFee function should emit ManagementFeeSet event.
304: Note, the setCap function should emit CapSet event.
559: Optimization, excessive assignment of depositReceipts[msg.sender].amount if condition is not met.
588: Optimization, the overflow check is not needed as the Solidity compiler 0.8+ performs it implicitly.
628: Major, adding the queuedWithdrawAmount to the locked balance for fee calculation is not always right. Consider a case last locked was 100 and queued 10, nothing changed in the current round, so current locked is 100 and queued is 10 again, the fee will be calculated from (100+10)-100 = 10. While there was no profit.
686: Major, the upgradeYearnVault() function will not work correctly as long as asset.balance(address(this)) > 0.**
_688: Note, instead of unwrapYieldToken call, should use simple collateral.withdraw._ - we need to know how much we are to unwrap
_689: Major, asset amount is expected instead of collateral amount._ this is fixed right?

YearnVaults\RibbonThetaYearnVault.sol
**289: Minor, the commitAndClose function could be executed twice resetting the lastLockedAmount to 0. This will result in fees taken from the whole amount.
423: Note, the unlockedAssedAmount variable has a typo and should be named unlockedAssetAmount.**

- depositYieldToken
- totalBalance
- mountToUnwrap >= stethBalance

### DIFF
--- a/contracts/vaults/BaseVaults/RibbonThetaVault.sol
+++ b/contracts/vaults/BaseVaults/RibbonThetaVault.sol
@@ -330,7 +330,7 @@ contract RibbonThetaVault is RibbonVault, RibbonThetaVaultStorage {
      */
     function _closeShort(address oldOption) private {
         uint256 lockedAmount = vaultState.lockedAmount;
-        if (optionState.currentOption != address(0)) {
+        if (oldOption != address(0)) {
             vaultState.lastLockedAmount = uint104(lockedAmount);
         }
         vaultState.lockedAmount = 0;
--- a/contracts/vaults/BaseVaults/RibbonThetaVault.sol
+++ b/contracts/vaults/BaseVaults/RibbonThetaVault.sol
@@ -329,11 +329,13 @@ contract RibbonThetaVault is RibbonVault, RibbonThetaVaultStorage {
      * @notice Closes the existing short position for the vault.
      */
     function _closeShort(address oldOption) private {
-        optionState.currentOption = address(0);
-
         uint256 lockedAmount = vaultState.lockedAmount;
-        vaultState.lastLockedAmount = uint104(lockedAmount);
+        if (optionState.currentOption != address(0)) {
+            vaultState.lastLockedAmount = uint104(lockedAmount);
+        }
         vaultState.lockedAmount = 0;
+
+        optionState.currentOption = address(0);
 
         if (oldOption != address(0)) {
             uint256 withdrawAmount =
@@ -397,14 +399,14 @@ contract RibbonThetaVault is RibbonVault, RibbonThetaVaultStorage {
      * @notice Burn the remaining oTokens left over from gnosis auction.
      */
     function burnRemainingOTokens() external onlyKeeper nonReentrant {
-        uint256 unlockedAssedAmount =
+        uint256 unlockedAssetAmount =
             VaultLifecycle.burnOtokens(
                 GAMMA_CONTROLLER,
                 optionState.currentOption
             );
 
         vaultState.lockedAmount = uint104(
-            uint256(vaultState.lockedAmount).sub(unlockedAssedAmount)
+            uint256(vaultState.lockedAmount).sub(unlockedAssetAmount)
         );
     }
 }
--- a/contracts/libraries/VaultLifecycle.sol
+++ b/contracts/libraries/VaultLifecycle.sol
@@ -251,10 +251,12 @@ library VaultLifecycle {
                 .div(oToken.strikePrice().mul(10**(10 + collateralDecimals)));
         } else {
             mintAmount = depositAmount;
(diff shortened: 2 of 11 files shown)


contracts/vaults/BaseVaults/RibbonThetaVault.sol
```diff
@@ -329,12 +329,14 @@ contract RibbonThetaVault is RibbonVault, RibbonThetaVaultStorage {
      * @notice Closes the existing short position for the vault.
      */
     function _closeShort(address oldOption) private {
-        optionState.currentOption = address(0);
-
         uint256 lockedAmount = vaultState.lockedAmount;
-        vaultState.lastLockedAmount = uint104(lockedAmount);
+        if (optionState.currentOption != address(0)) {
```
lastperson, line 333:
```suggestion
        if (oldOption != address(0)) {
```

lastperson: Could move into the if block below with the same condition.

kenchangh (approved):

PR #158, opened 2021-11-13
Ken/fix steth

kenchangh:


contracts/vaults/STETHVault/base/RibbonVault.sol
```diff
@@ -765,18 +760,21 @@ contract RibbonVault is
      * @return total balance of the vault, including the amounts locked in third party protocols
      */
     function totalBalance() public view returns (uint256) {
-        uint256 ethBalance =
-            IWETH(WETH).balanceOf(address(this)).add(address(this).balance);
```
kenchangh, line 769: Removes the WETH balance completely because we do not have mechanisms for unwrapping the WETH that we have received. This would cause an over-inflation of the pricePerShare as a result of counting the WETH balance

chuddster: I am confused...we do have the [mechanism](https://github.com/ribbon-finance/ribbon-v2/blob/7dd3dbdcfebd645e94a98247e7fb385c8ce86866/contracts/libraries/VaultLifecycleSTETH.sol#L420) to unwrap the WETH we have received. This is done in burnRemainingOTokens and rollToNextOption

PR #159, opened 2021-11-14
Fix fee collection

kenchangh:


contracts/vaults/BaseVaults/base/RibbonVault.sol
```diff
@@ -603,52 +607,23 @@ contract RibbonVault is
...
-        if (vaultFee > 0) {
-            transferAsset(payable(feeRecipient), vaultFee);
-            emit CollectVaultFees(
-                performanceFeeInAsset,
-                vaultFee,
-                vaultState.round,
-                feeRecipient
-            );
-        }
+        transferAsset(payable(recipient), totalVaultFee);
```
chuddster, line 624: we should probably make sure totalVaultFee > 0 before we transfer

contracts/libraries/VaultLifecycle.sol
```diff
@@ -144,32 +144,80 @@ library VaultLifecycle {
...
+
+        uint256 pricePerShareBeforeFee =
+            ShareMath.pricePerShare(
+                currentShareSupply,
+                currentBalance,
+                pendingAmount,
+                decimals
+            );
+
+        uint256 queuedWithdrawBeforeFee =
```
chuddster, line 195: should we have a ternary like we do here https://github.com/ribbon-finance/ribbon-v2/blob/853d281a7bc0fb67e18dd33f3eef5b0c33d479be/contracts/libraries/VaultLifecycle.sol#L237

contracts/vaults/YearnVaults/base/RibbonVault.sol
```diff
@@ -648,49 +654,21 @@ contract RibbonVault is
...
-                vaultState,
-                pastWeekBalance,
-                performanceFee,
-                managementFee
-            );
+        // Wrap entire `asset` balance to `collateralToken` balance
+        VaultLifecycleYearn.wrapToYieldToken(vaultParams.asset, collateral);
 
-        if (vaultFee > 0) {
-            VaultLifecycleYearn.withdrawYieldAndBaseToken(
```
chuddster, line 678: why are we not using withdrawYieldAndBaseToken?

chuddster (approved):

PR #160, opened 2021-11-15
Allow direct steth withdrawals

chuddster:


contracts/vaults/STETHVault/RibbonThetaSTETHVault.sol
```diff
@@ -269,16 +271,45 @@ contract RibbonThetaSTETHVault is RibbonVault, RibbonThetaSTETHVaultStorage {
 
         emit InstantWithdraw(msg.sender, amount, currentRound);
 
+        IERC20 steth = IERC20(collateralToken.stETH());
+        uint256 startStethBalance = steth.balanceOf(address(this));
+        bool enoughSTETH = true;
```
chuddster, line 276: enoughSTETH = (steth + wsteth) balance > withdraw amount. imagine someone deposits 1 eth and is only person in vault, and then wants to withdraw 1steth. this won't work so we will have to give them eth (since there is no steth balance or wsteth balance).

contracts/vaults/STETHVault/RibbonThetaSTETHVault.sol
```diff
@@ -269,16 +271,45 @@ contract RibbonThetaSTETHVault is RibbonVault, RibbonThetaSTETHVaultStorage {
...
+
+        if (withdrawYieldToken && amount > startStethBalance) {
```
chuddster, line 278: if we do need to withdraw, we check if there is enough even with wsteth unwrap and if so we do unwrap

contracts/vaults/STETHVault/base/RibbonVault.sol
```diff
@@ -484,22 +488,51 @@ contract RibbonVault is
                 vaultParams.decimals
             );
 
+        IERC20 steth = IERC20(collateralToken.stETH());
+        uint256 startStethBalance = steth.balanceOf(address(this));
+        bool enoughSTETH = true;
+
+        if (withdrawYieldToken && withdrawAmount > startStethBalance) {
```
chuddster, line 495: same logic as above

contracts/vaults/STETHVault/base/RibbonVault.sol
```diff
@@ -484,22 +488,51 @@ contract RibbonVault is
...
+            if (
+                withdrawAmount >
+                startStethBalance.add(
+                    collateralToken.getStETHByWstETH(
+                        collateralToken.balanceOf(address(this))
+                    )
```
chuddster, line 501: can consolidate into single vault lifecycle method for both completeWithdraw / withdrawInstantly + tests but now just checking logic

kenchangh:
Making it the default in: https://github.com/ribbon-finance/ribbon-v2/pull/214

PR #166, opened 2021-11-23
Rewrite unwrapYieldToken function

kenchangh:
The `unwrapYieldToken` function did not account for the scenario where we are 100% in stETH. This PR fixes it

### DIFF
--- a/contracts/libraries/VaultLifecycleSTETH.sol
+++ b/contracts/libraries/VaultLifecycleSTETH.sol
@@ -335,7 +335,7 @@ library VaultLifecycleSTETH {
      * @notice Unwraps the necessary amount of the wstETH token
      *         and transfers ETH amount to vault
      * @param amount is the amount of ETH to withdraw
-     * @param collateralToken is the address of the collateral token
+     * @param wstEth is the address of wstETH
      * @param stethToken is the address of stETH
      * @param crvPool is the address of the steth <-> eth pool on curve
      * @param minETHOut is the minimum eth amount to receive from the swap
@@ -344,7 +344,7 @@ library VaultLifecycleSTETH {
      */
     function unwrapYieldToken(
         uint256 amount,
-        address collateralToken,
+        address wstEth,
         address stethToken,
         address crvPool,
         uint256 minETHOut
@@ -353,10 +353,14 @@ library VaultLifecycleSTETH {
             amount >= minETHOut,
             "Amount withdrawn smaller than minETHOut from swap"
         );
+        require(
+            minETHOut.mul(10**18).div(amount) > 0.05 ether,
+            "Slippage on minETHOut too high"
+        );
 
-        uint256 ethAvailable = address(this).balance;
+        uint256 ethBalance = address(this).balance;
         IERC20 steth = IERC20(stethToken);
-        uint256 stethAvailable = steth.balanceOf(address(this));
+        uint256 stethBalance = steth.balanceOf(address(this));
 
         // 3 different success scenarios
         // Scenario 1. We hold enough ETH to satisfy withdrawal. Send it out directly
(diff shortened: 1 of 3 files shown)


contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -353,66 +353,74 @@ library VaultLifecycleSTETH {
...
+        }
+
+        // We unwrap if necessary first, then do the swap
+        uint256 ethstEthSum = ethAvailable.add(stethAvailable);
+        bool hasEnoughETHForWithdrawal = ethstEthSum >= amount.sub(1); // off by 1
+
+        // Scenario 3
+        {
+            if (!hasEnoughETHForWithdrawal) {
+                uint256 stethNeededFromUnwrap = amount.sub(ethstEthSum);
```
chuddster, line 377: should we do .sub(1) here

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -353,66 +353,74 @@ library VaultLifecycleSTETH {
...
+        // Scenario 2
+        // Now that we satisfied the ETH + stETH sum, we swap the stETH amounts necessary
+        // to facilitate a withdrawal
 
-        if (amountToUnwrap > 0) {
-            IWSTETH wsteth = IWSTETH(collateralToken);
-            IERC20 steth = IERC20(stethToken);
+        // This won't underflow since we already asserted that ethAvailable < amount before this
+        uint256 stEthAmountToSwap =
+            DSMath.min(amount.sub(ethAvailable), stethAvailable);
```
chuddster, line 401: I believe amount.sub(ethAvailable) will always be smaller

chuddster (approved):

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -353,66 +353,74 @@ library VaultLifecycleSTETH {
...
-        }
+        uint256 totalETHOut = ethAvailable.add(ethAmountOutFromSwap);
 
-        // This revert does not account for the ETH that is already unwrapped
         // Since minETHOut is derived from calling the Curve pool's getter,
         // it reverts in the worst case where the user needs to unwrap and sell
         // 100% of their ETH withdrawal amount
         require(
-            amountETHOut >= minETHOut,
+            totalETHOut >= minETHOut,
```
fqlx, line 419: If `minETHOut` out is set to 0 by the caller of this function, a MEV bot will hijack all the `stEth` during the curve swap.

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -353,66 +353,74 @@ library VaultLifecycleSTETH {
...
-                    .mul(wstETHPerStETH)
-                    .div(10**18);
-                // Unwrap to stETH
-                wsteth.unwrap(amountToUnwrap);
-            }
+        // CRV SWAP HERE from steth -> eth
+        // 0 = ETH, 1 = STETH
+        // We are setting 1, which is the smallest possible value for the _minAmountOut parameter
+        // However it is fine because we check that the totalETHOut >= minETHOut at the end
+        // which makes sandwich attacks not possible
```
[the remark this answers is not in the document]
kenchangh: This is missing the fact that the minETHOut is for the entire withdrawal amount, so for instance:

Vault holds = 5 ETH, 5 stETH
I initiate a withdrawal of amount=10 ETH, minETHOut=9.9 ETH

The curve swap would only do a swap of 5 stETH so it would always revert because the minETHOut is 9.9. So to cleanest solution here is to check at the end of the output

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -353,66 +353,74 @@ library VaultLifecycleSTETH {
...
+                        stethNeededFromUnwrap
+                    );
+
+                IWSTETH(collateralToken).unwrap(wstAmountToUnwrap);
+
+                // We do a final double check that we have enough to satisfy the amount
+                // We account for the off by 1 error
+                stethAvailable = steth.balanceOf(address(this));
+                require(
+                    ethAvailable.add(stethAvailable) >= amount.sub(2),
```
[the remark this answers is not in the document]
kenchangh: Fixed by using the minETHOut to check instead

contracts/libraries/VaultLifecycleSTETH.sol
kenchangh: Addressed by asserting that minETHOut has to be at least 0.95 * amount

fqlx (approved):

PR #178, opened 2021-12-03
Deploy fix for fee to account for withdrawals

kenchangh:
- Changed the upgrades test to be automatic. No need to manually update the storage values to test again, just read it off the contract, upgrade the proxy and check that it has not corrupted the state.
- Added a verifyContracts hardhat task


scripts/tasks/verifyContracts.ts
```diff
@@ -0,0 +1,78 @@
...
+    await run("verify:verify", {
+      address: RibbonThetaVaultSTETHLogic.address,
+      constructorArguments: THETA_VAULT_STETH_ARGS,
+      libraries: {
+        VaultLifecycle: VaultLifecycle.address,
+        VaultLifecycleSTETH: VaultLifecycleSTETH.address,
+      },
+    });
+  } catch (e) {}
+};
```
fqlx, line 77: Only four contracts are being verified here.  Do we plan on adding all contracts in this script?

test/upgrades/RibbonThetaVault.ts
```diff
@@ -75,161 +55,63 @@ describe("RibbonThetaVault upgrade", () => {
...
+    249,
+    250,
+    251,
+    252,
+    253,
+    254,
+    255,
+    256,
   ];
+  let storageLayout: [number | string, string][];
```
fqlx, line 101: Please avoid union types unless you have no other option, they cause more confusion than solve.

test/upgrades/RibbonThetaVault.ts
```diff
@@ -75,161 +55,63 @@ describe("RibbonThetaVault upgrade", () => {
...
-    lastStrikeOverrideRound:
-      "0x0000000000000000000000000000000000000000000000000000000000000000",
-    overriddenStrikePrice:
-      "0x0000000000000000000000000000000000000000000000000000000000000000",
-    auctionDuration:
-      "0x0000000000000000000000000000000000000000000000000000000000000e10",
-    optionAuctionID:
-      "0x000000000000000000000000000000000000000000000000000000000000003a",
-  });
+  checkIfStorageNotCorrupted("0x25751853Eab4D0eB3652B5eB6ecB102A2789644B");
```
fqlx, line 58: 1) Should this be the new address or the old address?

2) Could we define these addresses as like `RIBBON_THETA_VAULT_LOGIC_225751` or pull it from the deployments constants `constants/deployments.json` so we don't refine the same variable in multiple spots?

scripts/tasks/verifyContracts.ts
```diff
@@ -0,0 +1,78 @@
...
+  LDO_ADDRESS,
+  MARGIN_POOL,
+  OTOKEN_FACTORY,
+  STETH_ETH_CRV_POOL,
+  USDC_ADDRESS,
+  WETH_ADDRESS,
+  WSTETH_ADDRESS,
+} from "../../constants/constants";
+
+const main = async (_taskArgs: any, { deployments, network, run }) => {
```
fqlx, line 15: We should be linting no explicted `any`.

I think you can type tasks args like:
```
import { TaskArguments } from "hardhat/types";

taskArgs: TaskArguments
```

scripts/tasks/verifyContracts.ts
```diff
@@ -0,0 +1,78 @@
...
+  try {
+    await run("verify:verify", {
+      address: RibbonThetaVaultLogic.address,
+      constructorArguments: THETA_VAULT_ARGS,
+      libraries: { VaultLifecycle: VaultLifecycle.address },
+    });
+  } catch (e) {}
+
+  const THETA_VAULT_STETH_ARGS = [
+    WETH_ADDRESS[chainId],
```
chuddster, line 56: what about btc vault / aave vault

kenchangh: They are all under the ThetaVault contract

scripts/tasks/verifyContracts.ts
kenchangh: Possibly, but the other contracts can be automatically verified with the `yarn etherscan-verify` function

fqlx (approved):

PR #179, opened 2021-12-03
Add distribution and fee management for TreasuryVault

stevenwal:
Moved to a new PR


contracts/libraries/VaultLifecycleTreasury.sol
```diff
@@ -0,0 +1,796 @@
...
+    GammaTypes
+} from "../interfaces/GammaInterface.sol";
+import {IERC20Detailed} from "../interfaces/IERC20Detailed.sol";
+import {IGnosisAuction} from "../interfaces/IGnosisAuction.sol";
+import {SupportsNonCompliantERC20} from "./SupportsNonCompliantERC20.sol";
+import {UniswapRouter} from "./UniswapRouter.sol";
+
+library VaultLifecycleTreasury {
+    using SafeMath for uint256;
+    using SupportsNonCompliantERC20 for IERC20;
```
fqlx, line 23: The Open Zeppelin audit says this only supports USDT?
Here it looks like we are using this for collateral assets?

https://blog.openzeppelin.com/ribbon-finance-audit/

contracts/libraries/VaultLifecycleTreasury.sol
```diff
@@ -0,0 +1,796 @@
...
+            closeParams,
+            vaultParams,
+            underlying,
+            asset,
+            strikePrice,
+            expiry,
+            isPut
+        );
+
+        // get the black scholes premium of the option
```
fqlx, line 99: No need for this comment.  Maybe we need to rename the function to `getOptionPremium` or `getContractPrice`?

contracts/libraries/VaultLifecycleTreasury.sol
```diff
@@ -766,4 +786,158 @@ library VaultLifecycleTreasury {
...
+        }
+    }
+
+    function getNextMonthExpiry(uint256 timestamp, uint256 day) internal pure returns (uint256 nextMonthExpiry) {
+        uint secondsAccountedFor = 0;
+        uint buf;
+        uint8 i;
+        uint8 month;
+
+        // Year
```
fqlx, line 878: Usually when you have to add comments, it's a sign that it should be encapsulated into a private function.
Could we make this block something like `private getExpiryInSecondsForYear()`?

And then do this for seconds and months?

contracts/libraries/VaultLifecycleTreasury.sol
```diff
@@ -766,4 +786,158 @@ library VaultLifecycleTreasury {
...
+                + (period / 7) * 1 weeks;
+        }
+
+        nextExpiry = nextExpiry - (nextExpiry % (24 hours)) + (8 hours);
+        
+        return nextExpiry;
+    }
+
+    /************************************************
+     *  DATE HELPERS
```
fqlx, line 827: I'm a bit worried the date helpers could be buggy because dealing with date is a nightmare.  We really need to audit this and write tests, compare the date functions to another library, maybe in a different MR.

contracts/libraries/VaultLifecycleTreasury.sol
```diff
@@ -766,4 +786,158 @@ library VaultLifecycleTreasury {
...
+
+            // Year
+            year = uint16(ORIGIN_YEAR + timestamp / YEAR_IN_SECONDS);
+            numLeapYears = leapYearsBefore(year) - leapYearsBefore(ORIGIN_YEAR);
+
+            secondsAccountedFor += LEAP_YEAR_IN_SECONDS * numLeapYears;
+            secondsAccountedFor += YEAR_IN_SECONDS * (year - ORIGIN_YEAR - numLeapYears);
+
+            while (secondsAccountedFor > timestamp) {
+                    if (isLeapYear(uint16(year - 1))) {
```
fqlx, line 929:
```
leapYearsBefore -=  isLeapYear(uint16(year - 1)) ? LEAP_YEAR_IN_SECONDS : YEAR_IN_SECONDS
```

contracts/libraries/VaultLifecycleTreasury.sol
stevenwal: Yes, seems like. I'm not too sure about this

contracts/libraries/VaultLifecycleTreasury.sol
stevenwal: Im re-using this from existing VaultlifeCycle

contracts/libraries/VaultLifecycleTreasury.sol
stevenwal: Yes I agree, I'm actually referencing from here https://github.com/pipermerriam/ethereum-datetime/blob/master/contracts/DateTime.sol

contracts/libraries/VaultLifecycleTreasury.sol
stevenwal: The year, month and day are using a common variable which is secondsAccountedFor. I'm not entirely sure how this works but this variable is changed throughout to determine year, month and day

contracts/libraries/VaultLifecycleTreasury.sol
fqlx: Maybe we should import this as a library?

PR #180, opened 2021-12-03
Fixing lint:ts and adding no-explicit-any

fqlx:
- Linting all typescript files
- Added no-explicit-any
- Bumped node ci to v14


scripts/deploy/08_steth_call_theta_vault.ts
```diff
@@ -16,14 +11,17 @@ const main = async ({
   const chainId = network.config.chainId;
 
   if (chainId === CHAINID.AVAX_MAINNET || chainId === CHAINID.AVAX_FUJI) {
-    console.log(`08 - Skipping deployment stETH Call Theta Vault on ${network.name} because no stEth on Avax`);
+    console.log(
+      `08 - Skipping deployment stETH Call Theta Vault on ${network.name} because no stEth on Avax`
+    );
     return;
   }
```
chuddster, line 19: will these even deploy? we don't add AUCTION_DURATION, PREMIUM_DISCOUNT params in init when they are clearly there https://github.com/ribbon-finance/ribbon-v2/blob/600b44648a40aed5e3ca06c2700410e0dfba66d7/contracts/vaults/STETHVault/RibbonThetaSTETHVault.sol#L149

fqlx: I don't think so, the script is broken.  I didn't test this deploy script, only fixed the linting errors

fqlx: I added PREMIUM_DISCOUNT and AUCTION_DURATION back however I still think this script is failing and I already spent too much time trying to fix it on Kovan with no luck and don't really see this as a priority right now.

scripts/deploy/07_steth_logic.ts
```diff
@@ -11,6 +11,7 @@ import {
   STETH_ETH_CRV_POOL,
   WSTETH_ADDRESS,
 } from "../../constants/constants";
+import { PREMIUM_DISCOUNT, AUCTION_DURATION } from '../utils/constants';
```
chuddster, line 15: not adding to constructor, but to init. should be in 08_steth_call_theta_vault.ts

chuddster: should be right below here https://github.com/ribbon-finance/ribbon-v2/blob/4fcd7ba72c86c23d9e3081a005c2aa7a1226b787/scripts/deploy/08_steth_call_theta_vault.ts#L55

fqlx: Fixed

chuddster (approved):

PR #182, opened 2021-12-07
Fixing strike step on deploy script for avax/eth

fqlx:
- Fixing the strike step for Avax.  Should be 10 instead of 100 since Avax trades at ~$90 whereas ETH trades at ~$4000.


scripts/deploy/02_eth_call_theta_vault.ts
```diff
@@ -72,15 +73,23 @@ const main = async ({
   const strikeSelection = await deploy("StrikeSelectionETH", {
     contract: "StrikeSelection",
     from: deployer,
-    args: [pricer.address, STRIKE_DELTA, ETH_STRIKE_STEP],
+    args: [
+      pricer.address,
+      STRIKE_DELTA,
+      chainId === CHAINID.ETH_MAINNET ? ETH_STRIKE_STEP : AVAX_STRIKE_STEP,
```
kenchangh, line 79: We could follow how we declared `TOKEN_NAME` instead of doing this inline, ie

```
STRIKE_SELECTION_STEP = {[CHAINID.ETH_MAINNET]: ETH_STRIKE_STEP}
```

fqlx: Fixed

kenchangh (approved):

PR #185, opened 2021-12-08
Steven/ribbon treasuries/fee and distribution

stevenwal:
This is to add new capabilities allow the TreasuryVault to:
* distribute premium gained from options sale to whitelisted addresses
* charge performance fee in terms of premium asset (asset which denominates the options premium price)
* charge performance fee once auction is settled

TreasuryVault behaves similarly to ThetaVault except for the following changes:
* only whitelisted addresses can deposit
* interval between options sale is adjustable (selection includes biannually, quarterly, monthly, weekly, biweekly)
* auction is denominated in USDC by default
* management fee is charged regardless of performance
* performance fee is charged in USDC by default


contracts/libraries/DateTime.sol
```diff
@@ -0,0 +1,271 @@
...
+            month == 8 ||
+            month == 10 ||
+            month == 12
+        ) {
+            daysInMonth = 31;
+        } else if (month != 2) {
+            daysInMonth = 30;
+        } else {
+            daysInMonth = _isLeapYear(year) ? 29 : 28;
+        }
```
kenchangh, line 146: You can simplify these if-else and save gas by doing

```solidity
if (month % 2 != 0) { // replicates the odd-number months
  
} else if (month != 2) {
 
} else {

}
```

contracts/libraries/VaultLifecycleTreasury.sol
```diff
@@ -225,19 +255,18 @@ library VaultLifecycleTreasury {
                 .add(withdrawAmountDiff);
         }
 
-        {
-            (performanceFeeInAsset, , totalVaultFee) = getVaultFees(
+        // We do not charge on the first round
+        managementFeeInAsset = vaultState.round <= 1
```
kenchangh, line 259: Instead of adding this `vaultState.round <= 1` check that will run on all subsequent rounds, we could instead pass in `params.managementFee = 0`. This will help us avoid looking up `vaultState.round` from storage too

contracts/vaults/TreasuryVault/RibbonTreasuryVault.sol
```diff
@@ -797,25 +752,22 @@ contract RibbonTreasuryVault is
...
                 newPricePerShare,
                 mintShares,
-                performanceFeeInAsset,
-                totalVaultFee
-            ) = VaultLifecycle.rollover(
+                managementFeeInAsset
+            ) = VaultLifecycleTreasury.rollover(
                 vaultState,
-                VaultLifecycle.RolloverParams(
+                VaultLifecycleTreasury.RolloverParams(
```
kenchangh, line 766: Adding on to this: https://github.com/ribbon-finance/ribbon-v2/pull/185/files#r767468586

Since we access the `uint currentRound = vaultState.round;`, we can do:

```solidity
...
lastQueuedWithdrawAmount,
currentRound != 1 ? managementFee : 0)
```

contracts/vaults/TreasuryVault/RibbonTreasuryVault.sol
```diff
@@ -369,9 +313,20 @@ contract RibbonTreasuryVault is
      */
     function setCap(uint256 newCap) external onlyOwner {
         require(newCap > 0, "!newCap");
+
+        uint256 previousCap = vaultParams.cap;
+        vaultParams.cap = _getCap104(newCap);
```
kenchangh, line 318: I don't really see a good reason refactoring this out into a separate function imo. Makes it hard to read

contracts/vaults/TreasuryVault/RibbonTreasuryVault.sol
```diff
@@ -339,13 +267,29 @@ contract RibbonTreasuryVault is
...
+        returns (uint256)
+    {
+        uint256 _period = period;
+        uint256 feeDivider =
+            _period % 30 == 0
+                ? Vault.FEE_MULTIPLIER.mul(12 / (_period / 30))
+                : WEEKS_PER_YEAR.div(_period / 7);
+
+        // We are dividing annualized management fee by num weeks in a year
+        return managementFee.mul(Vault.FEE_MULTIPLIER).div(feeDivider);
```
kenchangh, line 292: We seem to be mixing between using `mul` `div` and normal signs. We should standardize them by doing `*` `/`, since the overflow checks are already done for us in Solidity 0.8

contracts/storage/RibbonTreasuryVaultStorage.sol
```diff
@@ -70,6 +67,9 @@ abstract contract RibbonTreasuryVaultStorageV1 {
     /// @notice Amount locked for scheduled withdrawals last week;
     uint256 public lastQueuedWithdrawAmount;
 
+    /// @notice Period between each options sale.
```
kenchangh, line 70: Can add that it can only be 7, 14, ,30, 90, 180

contracts/vaults/TreasuryVault/RibbonTreasuryVault.sol
```diff
@@ -1000,10 +953,80 @@ contract RibbonTreasuryVault is
...
+            recipient
+        );
+    }
+
+    /**
+     * @notice Distribute the premium to whitelisted addresses
      */
-    function settleAuction() external onlyKeeper nonReentrant {
-        VaultLifecycle.settleAuction(GNOSIS_EASY_AUCTION, optionAuctionID);
+    function _distribute(IERC20 token, uint256 amount) internal {
```
kenchangh, line 1017: nit _distributePremium?

contracts/vaults/TreasuryVault/RibbonTreasuryVault.sol
```diff
@@ -1000,10 +953,80 @@ contract RibbonTreasuryVault is
...
+        // Distribute to whitelisted address
+        address[] memory _whitelist = whitelistArray;
```
kenchangh, line 1019: It's cheaper if you declare this as `storage`

contracts/vaults/TreasuryVault/RibbonTreasuryVault.sol
```diff
@@ -1000,10 +953,80 @@ contract RibbonTreasuryVault is
...
+    }
+
+    /**
+     * @notice Charge performance fee and distribute remaining to whitelisted address
+     */
+    function chargeAndDistribute() external onlyKeeper nonReentrant {
+        require(
+            IERC20(USDC).balanceOf(address(this)) > 0,
+            "no premium to distribute"
+        );
```
kenchangh, line 974: Shouldn't this check go into `_chargeAndDistribute`?

contracts/vaults/TreasuryVault/RibbonTreasuryVault.sol
```diff
@@ -868,16 +813,25 @@ contract RibbonTreasuryVault is
      *         This allows all the users to withdraw if the next option is malicious.
      */
     function commitAndClose() external nonReentrant {
+        uint256 stableBalance = IERC20(USDC).balanceOf(address(this));
+
+        // In case chargeAndDistribute was not called last round, call
+        // the function to conclude last round's performance fee and distribution
+        if (stableBalance > 0) {
+            _chargeAndDistribute();
+        }
```
kenchangh, line 822: We should generally do token transfers at the end of the function

kenchangh: See: https://fravoll.github.io/solidity-patterns/checks_effects_interactions.html

contracts/vaults/TreasuryVault/RibbonTreasuryVault.sol
```diff
@@ -339,13 +267,29 @@ contract RibbonTreasuryVault is
...
+     */
+    function _perRoundManagementFee(uint256 managementFee)
+        internal
+        view
+        returns (uint256)
+    {
+        uint256 _period = period;
+        uint256 feeDivider =
+            _period % 30 == 0
+                ? Vault.FEE_MULTIPLIER.mul(12 / (_period / 30))
```
kenchangh, line 288: Why do we need to multiply by `Vault.FEE_MULTIPLIER` here? I see we are already multiplying it on line 292

contracts/vaults/TreasuryVault/RibbonTreasuryVault.sol
stevenwal: This was to prevent double checks. in `commitAndClose` we are already checking balance before calling `_chargeAndDistribute`.

contracts/vaults/TreasuryVault/RibbonTreasuryVault.sol
stevenwal: This is to follow the `WEEKS_PER_YEAR` convention.

```Number of weeks per year = 52.142857 weeks * FEE_MULTIPLIER = 52142857```

For the monthly, quarterly and biannually, we need to do the same. So if we have monthly, the management fee will be divided by 12000000.

contracts/vaults/TreasuryVault/RibbonTreasuryVault.sol
stevenwal: Added this guard to `_chargeAndDistribute`. For functions where `_chargeAndDistribute` is optional (e.g. `concludeOptionsSale` and `commitAndClose`, the function will check balance and only call `_chargeAndDistribute` when balance is larger than 0.

kenchangh (approved):

PR #189, opened 2021-12-13
[WIP] Hardhat fork caching to speed up CI

kenchangh:
It takes a few minutes for hardhat tests to start up because it needs to load the state over the network. This caches the fork state so that it improves test speeds


kenchangh:
Stale

PR #194, opened 2021-12-21
feat/Add stake() to RibbonThetaVault to allow staking Vault shares in 1 tx

0xTomoyo:
- Adds a `liquidityGauge` variable to the `RibbonThetaVault` contract and a `setLiquidityGauge` function
- `stake()` redeems shares, transfers it back to the vault, approves it to `liquidityGauge` & then stakes it on behalf of the sender


contracts/tests/MockStakingRewards.sol
```diff
@@ -0,0 +1,52 @@
...
+    constructor(address _stakingToken) {
+        require(
+            _stakingToken != address(0),
+            "Staking token must be non-zero address"
+        );
+        stakingToken = IERC20(_stakingToken);
+    }
+
+    function totalSupply() external view returns (uint256) {
+        return _totalSupply;
```
fqlx, line 28: I don't really see a reason to make `_totalSupply` private and then have a function to return it.

contracts/vaults/BaseVaults/RibbonThetaVault.sol
```diff
@@ -322,6 +335,19 @@ contract RibbonThetaVault is RibbonVault, RibbonThetaVaultStorage {
         );
     }
 
+    /**
+     * @notice Stakes a users vault shares
+     * @param numShares is the number of shares to stake
+     */
+    function stake(uint256 numShares) external nonReentrant {
```
chuddster, line 342: Imagine Alice has 50 shares of rETH-THETA. She does a redeemMax() for 50 shares. She then calls `stake(50 shares)`. 

This will revert because we are trying to redeem shares that have already been redeemed, but Alice staking her separately redeemed shares should be a valid flow.  

I recommend replacing `_redeem(numShares, false);` with 

```
(uint256 heldByAccount, uint256 heldByVault) = shareBalances(msg.sender); 
require(numShares <= heldByAccount.add(heldByVault), "numShares greater than shares owned!");
if(heldByAccount < numShares){
      _redeem(numShares.sub(heldByAccount), false);
}
```

test/RibbonThetaVault.ts
```diff
@@ -3322,6 +3322,122 @@ function behavesLikeRibbonOptionsVault(params: {
...
+        await assetContract
+          .connect(userSigner)
+          .approve(vault.address, depositAmount);
+        await vault.deposit(depositAmount);
+        await rollToNextOption();
+        await expect(vault.stake(depositAmount)).to.be.revertedWith(
+          "!stakingRewards"
+        );
+      });
+
```
chuddster, line 3346: you should add a test case for 1. deposit 2. maxRedeem 3. stake(prev redeemed shares)

contracts/tests/MockStakingRewards.sol
0xTomoyo: Fixed here: https://github.com/ribbon-finance/ribbon-v2/pull/194/commits/92bfe9b03ee5c687cb20a46e179a713040d6aee6

contracts/vaults/BaseVaults/RibbonThetaVault.sol
0xTomoyo: good idea, just added it here (along with tests for it): https://github.com/ribbon-finance/ribbon-v2/pull/194/commits/92bfe9b03ee5c687cb20a46e179a713040d6aee6
I didn't add the require line since it would anyways revert when calling `_transfer` if numShares were to exceed the shares the user owned

test/RibbonThetaVault.ts
0xTomoyo: Added: https://github.com/ribbon-finance/ribbon-v2/pull/194/commits/92bfe9b03ee5c687cb20a46e179a713040d6aee6

0xTomoyo:
`RibbonThetaVault` is now really close to the 24kb contract size limit after adding the `stake()` function
Screen Shot 2021-12-22 at 4 51 20 pm
For now I haven't added revert messages to `stake()` (adding them pushed the contract size over the limit)
Screen Shot 2021-12-22 at 4 52 41 pm

contracts/vaults/BaseVaults/RibbonThetaVault.sol
```diff
@@ -322,6 +335,22 @@ contract RibbonThetaVault is RibbonVault, RibbonThetaVaultStorage {
...
+        require(stakingRewards != address(0)); // Removed revert msgs due to contract size limit
+        require(numShares > 0);
+        (uint256 heldByAccount, ) = shareBalances(msg.sender);
```
kenchangh, line 345: If we're only interested in `heldByAccount`, this is equivalent but with less gas use:

```
uint256 heldByAccount = balanceOf(msg.sender);
```

0xTomoyo: Oh nice, just changed it now

contracts/vaults/BaseVaults/RibbonThetaVault.sol
```diff
@@ -322,6 +335,22 @@ contract RibbonThetaVault is RibbonVault, RibbonThetaVaultStorage {
...
+     */
+    function stake(uint256 numShares) external nonReentrant {
+        require(stakingRewards != address(0)); // Removed revert msgs due to contract size limit
+        require(numShares > 0);
+        uint256 heldByAccount = balanceOf(msg.sender);
+        if (heldByAccount < numShares) {
+            _redeem(numShares.sub(heldByAccount), false);
+        }
+        _transfer(msg.sender, address(this), numShares);
+        _approve(address(this), stakingRewards, numShares);
```
kenchangh, line 350: Instead of doing an intermediate `_transfer` step to `address(this)`, we could approve on behalf of the user, eg:

```solidity
_approve(msg.sender, stakingRewards, numShares);
IStakingRewards(stakingRewards).stakeFor(numShares, msg.sender);

// for security after stakeFor, we can verify that the stakingRewards behaves as it should be
require(balanceOf(stakingRewards).sub(startStakingRewardsBalance) == numShares, "Wrong amount staked");
```

contracts/vaults/BaseVaults/RibbonThetaVault.sol
```diff
@@ -322,6 +335,22 @@ contract RibbonThetaVault is RibbonVault, RibbonThetaVaultStorage {
         );
     }
 
+    /**
+     * @notice Stakes a users vault shares
+     * @param numShares is the number of shares to stake
+     */
+    function stake(uint256 numShares) external nonReentrant {
+        require(stakingRewards != address(0)); // Removed revert msgs due to contract size limit
```
kenchangh, line 343: Can cache `stakingRewards` since it's accessed twice

0xTomoyo: Just added caching: https://github.com/ribbon-finance/ribbon-v2/pull/194/commits/b922790fc59b18311f214817bb237acb1b88d9b3

contracts/vaults/BaseVaults/RibbonThetaVault.sol
```diff
@@ -282,6 +283,18 @@ contract RibbonThetaVault is RibbonVault, RibbonThetaVaultStorage {
...
     }
 
+    /**
+     * @notice Sets the new stakingRewards contract for this vault
+     * @param newStakingRewards is the address of the new stakingRewards contract
+     */
+    function setStakingRewards(address newStakingRewards)
+        external
+        onlyOwner
+        nonReentrant
```
kenchangh, line 293: Hack to reduce bytecode size: since `nonReentrant` is a modifier that gets copied, we can safely remove it for this function because we assume the owner will not call it and cause reentrancy

0xTomoyo: Removed here: https://github.com/ribbon-finance/ribbon-v2/pull/194/commits/8a823b5a3d08f85388002a012fec7db7adfa475f

kenchangh (approved):

kenchangh (approved):

PR #195, opened 2021-12-21
Steven/ribbon treasury fix

stevenwal:
This is to make fixes to the existing RibbonTreasuryVault:
1. Fixed a mistake in getNextExpiry where the biannual expiry is calling quarterly expiry function
2. Added new events (InitiateGnosisAuction) and (DistributePremium) for subgraph purposes


contracts/vaults/TreasuryVault/RibbonTreasuryVault.sol
```diff
@@ -1005,16 +1019,23 @@ contract RibbonTreasuryVault is
...
         for (uint256 i = 0; i < _whitelist.length; i++) {
             // Distribute to whitelist proportional to the amount of
             // shares they own
-            token.safeTransfer(
-                _whitelist[i],
-                shares(_whitelist[i]).mul(amount).div(totalSupply)
-            );
+            _amounts[i] = shares(_whitelist[i]).mul(amount).div(totalSupply);
+
+            token.safeTransfer(_whitelist[i], _amounts[i]);
```
kenchangh, line 1030: Accessing _whitelist[i] twice here so we can cache it in a variable. Alternatively we can keep to how we were writing before

contracts/vaults/TreasuryVault/RibbonTreasuryVault.sol
```diff
@@ -1005,16 +1019,23 @@ contract RibbonTreasuryVault is
...
         }
+
+        emit DistributePremium(
+            amount,
+            _amounts,
+            _whitelist,
```
kenchangh, line 1036: I'm not sure this works because `_whitelist` is a storage reference

contracts/vaults/TreasuryVault/RibbonTreasuryVault.sol
stevenwal: I'm caching it in a variable, reason is to store the distributed amount into the array for easier subgraph mapping.

contracts/vaults/TreasuryVault/RibbonTreasuryVault.sol
stevenwal: This works, for now

kenchangh (approved):

PR #199, opened 2021-12-23
Resolve options expiry to next friday

kenchangh:
Currently, we run into a situation where we are not able to roll the funds if you do not call `commitAndClose` 1 week within the options expiry. This changes it so that you could do so.


stevenwal (approved):
Looks good

PR #200, opened 2021-12-23
Testing public variables on upgrades instead of storage slots

kenchangh:
- The previous upgrade tests were wrong because they looked at the storage slots
- The correct approach would be to lookup the values using the public variables, because on upgrades the implementation change would change where the variable was pointing to
- Tested by deliberately breaking the order of the variables in RibbonThetaVaultStorage

### DIFF
--- a/test/helpers/utils.ts
+++ b/test/helpers/utils.ts
@@ -17,6 +17,7 @@ import {
 import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/dist/src/signers";
 import { BigNumber, BigNumberish, Contract } from "ethers";
 import { wmul } from "../helpers/math";
+import { string } from "hardhat/internal/core/params/argumentTypes";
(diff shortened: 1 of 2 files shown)


0xTomoyo (approved):

test/helpers/utils.ts
```diff
@@ -446,3 +446,38 @@ export function encodePath(tokenAddresses, fees) {
 
   return encoded.toLowerCase();
 }
+
+/* eslint @typescript-eslint/no-explicit-any: "off" */
+export const objectEquals = (a: any, b: any) => {
```
fqlx, line 451: nit: I'd prefer if we used lodash for common utility functions rather than copying functions into the `utils.ts`

https://lodash.com/docs/4.17.15#isEqual

fqlx (approved):

PR #203, opened 2021-12-28
Change whitelist method to depositors limit

stevenwal:
Changes include:
- remove all whitelist related variables and methods
- put a limit on the number of depositors in the vault
- add method to store depositors address in array and mapping during deposit
- add method to remove depositors address when depositor has no pending deposit and shares


contracts/vaults/TreasuryVault/RibbonTreasuryVault.sol
```diff
@@ -75,8 +75,8 @@ contract RibbonTreasuryVault is
     // The minimum duration for an option auction.
     uint256 private constant MIN_AUCTION_DURATION = 5 minutes;
 
-    // Maximum number of whitelisted user address.
-    uint256 private constant WHITELIST_LIMIT = 5;
+    // Maximum number of depositor user address.
+    uint256 private constant DEPOSITORS_LIMIT = 30;
```
kenchangh, line 79: Let's make this configurable by the owner, and public

contracts/vaults/TreasuryVault/RibbonTreasuryVault.sol
```diff
@@ -464,11 +435,15 @@ contract RibbonTreasuryVault is
...
      * @param amount is the amount of `asset` to deposit
      */
-    function deposit(uint256 amount) external onlyWhitelist nonReentrant {
+    function deposit(uint256 amount) external nonReentrant {
         require(amount > 0, "!amount");
 
         _depositFor(amount, msg.sender);
 
+        // if (!depositorsMap[msg.sender]) {
+        _addDepositor(msg.sender);
```
kenchangh, line 444: We can move this to before `_depositFor` so that it throws when they exceed the depositor limit

contracts/vaults/TreasuryVault/RibbonTreasuryVault.sol
stevenwal: Moved this variable to TreasuryStorage.

contracts/vaults/TreasuryVault/RibbonTreasuryVault.sol
```diff
@@ -404,69 +389,68 @@ contract RibbonTreasuryVault is
     }
 
     /**
-     * @notice Adds new whitelisted address
-     * @param newWhitelist is the address to include in the whitelist
+     * @notice Set the maximum number of depositors
+     * @param newMaxDepositors is the new cap for number of depositors
      */
-    function addWhitelist(address newWhitelist)
+    function setMaxDepositors(uint128 newMaxDepositors)
```
kenchangh, line 395: We can make this uint256

kenchangh (approved):

PR #209, opened 2022-01-24
Peckshield audit fixes for Ribbon Treasury

stevenwal:
Peckshield audit fixes for Ribbon Treasury


stevenwal:
**PVE-004**
In initialization, the contract sets the variable `vaultState.lastLockedAmount`. [Reference](https://github.com/ribbon-finance/ribbon-v2/blob/master/contracts/vaults/TreasuryVault/RibbonTreasuryVault.sol#L227).

This is different from the one we used for `totalBalance()` which is `vaultState.lockedAmount`. [Reference](https://github.com/ribbon-finance/ribbon-v2/blob/master/contracts/vaults/TreasuryVault/RibbonTreasuryVault.sol#L1111)

Hence, there is no double counting here.

kenchangh (approved):

PR #210, opened 2022-01-28
WIP: feat/avax-mim Adding AVAX MIM Puts

fqlx:
- feat/avax-mim Adding AVAX MIM Puts
- Allowing us to use a different strike and collateral besides USDC


test/RibbonThetaVault.ts
```diff
@@ -176,6 +183,40 @@ describe("RibbonThetaVault", () => {
     ],
   });
 
+  behavesLikeRibbonOptionsVault({
+    name: `Ribbon AVAX MIM Theta Vault (Put)`,
+    tokenName: "Ribbon AVAX MIM Theta Vault Put",
+    tokenSymbol: "rMIM-THETA-P",
```
chuddster, line 189: should be rETH-THETA-P-MIM probably.

PR #214, opened 2022-02-04
Withdraw stETH directly

0xTomoyo:
Forces all withdrawals to receive stETH (converts wstETH/ETH to stETH then sends it to the user)


contracts/vaults/STETHVault/base/RibbonVault.sol
```diff
@@ -480,25 +479,20 @@ contract RibbonVault is
...
+        IERC20(STETH).safeTransfer(
+            msg.sender,
+            VaultLifecycleSTETH.withdrawStEth(
                 STETH,
-                STETH_ETH_CRV_POOL,
-                minETHOut
-            );
+                address(collateralToken),
+                withdrawAmount
+            )
```
kenchangh, line 488: May be redundant, but to be safe, we can add a `require` that checks that the amount returned by `withdrawStEth` is equal to the withdrawAmount

```solidity
uint stEthWithdrawAmount = VaultLifecycleSTETH.withdrawStEth(
    STETH,
    address(collateralToken),
    withdrawAmount
);
require(stEthWithdrawAmount == withdrawAmount)
```

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -499,6 +499,62 @@ library VaultLifecycleSTETH {
...
+            // Can send out the stETH directly
+            return amount; // We return here if we have enough stETH to satisfy the withdrawal
+        } else {
+            // If amount > stethBalance, send out the entire stethBalance and check wstETH and ETH
+            amount = amount.sub(stethBalance);
+        }
+        uint256 wstethBalance = IWSTETH(wstEth).balanceOf(address(this));
+        stethBalance = IWSTETH(wstEth).getStETHByWstETH(wstethBalance);
+        if (stethBalance >= amount) {
+            wstethBalance = IWSTETH(wstEth).getWstETHByStETH(amount);
```
kenchangh, line 529: My understanding from reading the steth contracts are that it's these functions that are adding to the rounding errors. Each conversion rounds up the amounts (muldiv rounded up), so when it's `.add(3)` in the tests, it probably has gone through:

1. getStETHByWstETH
2. getWstETHByStETH
3. unwrap

One way to save gas and lessen this rounding error is to do the muldiv ourselves, eg:

```solidity
uint stEthPerToken = IWSTETH(wstEth).stEthPerToken();
stethBalance = (wstethBalance * stEthPerToken)/10**18; // this rounds down by default
wstethBalance = (amount * 10**18) / stEthPerToken;

IWSTETH(wstEth).unwrap(wstethBalance);
```

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -499,6 +499,62 @@ library VaultLifecycleSTETH {
...
+                IERC20(steth).balanceOf(address(this)).sub(stethBalance)
+            );
+        }
+        // Wrap ETH to stETH if we don't have enough stETH + wstETH
+        uint256 ethBalance = address(this).balance;
+        if (amount > 0 && ethBalance >= amount) {
+            ISTETH(steth).submit{value: amount}(address(this));
+        } else if (ethBalance > 0) {
+            ISTETH(steth).submit{value: ethBalance}(address(this));
+        }
```
kenchangh, line 552: Should we add an `else` statement here and revert?

contracts/libraries/VaultLifecycleSTETH.sol
0xTomoyo: Good idea, `stEthPerToken` saves gas
In the background it makes an additional call to the stETH, so its cheaper to get the `getTotalShares` and `getTotalPooledEther` directly from stETH and do the calculations

```
uint256 totalShares = ISTETH(steth).getTotalShares();
uint256 totalPooledEther = ISTETH(steth).getTotalPooledEther();
stethBalance = wstethBalance.mul(totalPooledEther).div(totalShares);

wstethBalance = amount.mul(totalShares).div(totalPooledEther);
```

contracts/vaults/STETHVault/base/RibbonVault.sol
0xTomoyo:
```
// Accounts for rounding errors by a margin of 3 wei
require(_amount.add(3) >= stethBalance, "Unwrapped too much stETH");
require(_amount <= stethBalance.add(3), "Unwrapped insufficient stETH");
return stethBalance; // We return here if we have enough stETH + wstETH + ETH
```

Added this to `withdrawStEth` to account for rounding errors
Although doing the muldiv ourselves lessens to the rounding errors, there still are errors when calling `unwrap` on wstETH and `submit` on stETH
In my tests a margin of 2 wei had issues when withdrawing both stETH + wstETH and stETH + ETH, so I set it to 3 wei which worked for all conditions.

contracts/libraries/VaultLifecycleSTETH.sol
0xTomoyo: I think having this require check is enough: https://github.com/ribbon-finance/ribbon-v2/pull/214#discussion_r805522611

kenchangh (approved):

PR #223, opened 2022-02-22
WIP: task/savax santity check

fqlx:
WIP: task/savax santity check


contracts/oracles/SAvaxOracle.sol
```diff
@@ -47,6 +46,6 @@ contract SAvaxOracle is IPriceOracle {
     }
 
     function decimals() external view override returns (uint256) {
-        return sAVAX.decimals();
+        return WAVAXOracle.decimals();
```
kenchangh, line 49: We can just hardcode to 8 here right?

fqlx: Yeah I could but I think it's better to reference the underlying so we know why it's hardcoded to 8

kenchangh (approved):

PR #226, opened 2022-02-28
feat/SAVAXDepositHelper Adding sAVAX deposit helper so users can stake AVAX and deposit on Ribbon

fqlx:
feat/SAVAXDepositHelper Adding sAVAX deposit helper so users can stake AVAX and deposit on Ribbon

https://snowtrace.io/address/0x084a82fe209f014b3cb6D5986B2C65B584168303


contracts/interfaces/IRibbonVault.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+pragma solidity =0.8.4;
+
+import {Vault} from "../libraries/Vault.sol";
+
+interface IRibbonVault {
```
kenchangh, line 6: We could add IRibbonVault into IRibbon.sol instead to centralize the interfaces

contracts/interfaces/ISAVAX.sol
```diff
@@ -1,7 +1,11 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.4;
 
-interface ISAVAX {
+import {
+    IERC20Upgradeable
+} from "@openzeppelin/contracts-upgradeable/token/ERC20/IERC20Upgradeable.sol";
+
+interface ISAVAX is IERC20Upgradeable {
```
kenchangh, line 8: Does it it need to be IERC20Upgradeable? My understanding is that it is a plain IERC20

contracts/utils/SAVAXDepositHelper.sol
```diff
@@ -0,0 +1,24 @@
...
+    constructor(address _sAVAX, address _sAVAXVault) {
+        require(_sAVAX != address(0), "!sAVAX");
+        require(_sAVAXVault != address(0), "!sAVAXVault");
+
+        sAVAX = ISAVAX(_sAVAX);
+        sAVAXVault = IRibbonVault(_sAVAXVault);
+    }
+
+    function deposit() external payable {
+        uint256 sAVAXAmount = sAVAX.submit{value: msg.value}();
```
kenchangh, line 20: Do we have to cover the edge case where `sAVAXAmount` is not equal to `msg.value` here? Since the user sends in the `msg.value` there may be cases where it is larger than `sAVAXAmount` and so some dust gets stuck in the contract.

We'll have to do a refund in that scenario:

```sol
uint256 sAVAXAmount = sAVAX.submit{value: msg.value}();

if (msg.value > sAVAXAmount) {
  (bool success, ) = msg.sender.call{value:msg.value-sAVAXAmount}("");
  require(success, "Transfer failed.");
}
```

test/utils/SAVAXDepositHelper.ts
```diff
@@ -0,0 +1,56 @@
...
+  let signer: SignerWithAddress;
+
+  before(async () => {
+    await network.provider.request({
+      method: "hardhat_reset",
+      params: [
+        {
+          forking: {
+            jsonRpcUrl: process.env.AVAX_URI,
+            blockNumber: 11518500,
```
kenchangh, line 22: I don't see a similar block number forked in the existing tests, would increase test speed if we could share the same block numbers :)

contracts/utils/SAVAXDepositHelper.sol
fqlx: `sAVAXAmount` is less than `msg.value`, never equal or greater.
I suppose we could add a check? But I dunno if this is required

test/utils/SAVAXDepositHelper.ts
fqlx: Yeah - will fix but it will break the other tests unless the blocknumber >1 day.

contracts/interfaces/ISAVAX.sol
fqlx: It's a proxy:
https://snowtrace.io/address/0x2b2c81e08f1af8835a78bb2a90ae924ace0ea4be#code

Here it's `IERC20Upgradeable` (line 1346):

```
contract StakedAvax is
    IERC20Upgradeable,
    AccessControlUpgradeable,
    ReentrancyGuardUpgradeable,
    PausableUpgradeable,
    StakedAvaxStorage
{
```

test/utils/SAVAXDepositHelper.ts
```diff
@@ -0,0 +1,56 @@
...
+      (await sAVAXVault.depositReceipts(signer.address))[AMOUNT_INDEX].toString(),
+      "0"
+    );
+
+    await sAVAXDepositHelper.deposit({ value: ethers.utils.parseEther('.26') });
+
+    assert.equal(
+      (await sAVAXVault.depositReceipts(signer.address))[AMOUNT_INDEX].toString(),
+      "259480652035615045"
+    );
```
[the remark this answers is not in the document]
fqlx: Added

contracts/utils/SAVAXDepositHelper.sol
fqlx: Maybe if their validator gets slashed too many times? `sAVAXAmount` could be greater than `msg.value`?  I don't really think these scenarios will happen.

contracts/interfaces/IRibbonVault.sol
fqlx: We have another file `IRibbonThetaVault.sol`.  Do we move that as well or should I delete it?

contracts/utils/SAVAXDepositHelper.sol
shibacrypt: sAVAX is not 1:1 with AVAX, its a share-based receipt so sAVAX will always be < msg.value (AVAX amount deposited).

contracts/interfaces/ISAVAX.sol
shibacrypt: there's no need to extend `IERC20Upgradeable` , in fact we only need this function for the sAVAX-specific transactions

```
    function submit() external payable returns (uint256);
```

contracts/utils/SAVAXDepositHelper.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: MIT
+pragma solidity =0.8.4;
+
+import {ISAVAX} from "../interfaces/ISAVAX.sol";
+import {IRibbonVault} from "../interfaces/IRibbonVault.sol";
+
+contract SAVAXDepositHelper {
+    ISAVAX public sAVAX;
+    IRibbonVault public sAVAXVault;
```
shibacrypt, line 9: Nit: Could be `immutable`

contracts/utils/SAVAXDepositHelper.sol
```diff
@@ -0,0 +1,24 @@
...
+        require(_sAVAX != address(0), "!sAVAX");
+        require(_sAVAXVault != address(0), "!sAVAXVault");
+
+        sAVAX = ISAVAX(_sAVAX);
+        sAVAXVault = IRibbonVault(_sAVAXVault);
+    }
+
+    function deposit() external payable {
+        uint256 sAVAXAmount = sAVAX.submit{value: msg.value}();
+        sAVAX.approve(address(sAVAXVault), sAVAXAmount);
```
shibacrypt, line 21: sAVAX's interface would need this, consider adding

```
    function approve(address _recipient, uint256 _amount) external returns (bool);
```

to the `import {ISAVAX} from "../interfaces/ISAVAX.sol";` interface

contracts/interfaces/ISAVAX.sol
shibacrypt: Seems like we need an `approve` method here: https://github.com/ribbon-finance/ribbon-v2/pull/226/files#r816417713

fqlx: @shibacrypt `approve()` is inferred from the `IERC20Upgradeable` interface

contracts/utils/SAVAXDepositHelper.sol
fqlx: `approve()`'s definition will be inherited from `IERC20` so no need to add

contracts/interfaces/IRibbonVault.sol
fqlx: Done

test/utils/SAVAXDepositHelper.ts
fqlx: Fixed

kenchangh (approved):
Let's go 🚢

kenchangh (dismissed):

stevenwal (approved):

PR #227, opened 2022-03-03
Add vol param for getStrikePrice

kenchangh:
- Added a helper function getStrikePriceWithVol which helps frontends to get the strike price at a specific vol


stevenwal (approved):
LGTM

PR #229, opened 2022-03-04
Make setPremiumDiscount onlyKeeper

kenchangh:
- Changes setPremiumDiscount to onlyKeeper rather than onlyOwner
- Required to adjust the min prices for the auction so we do not give a massive discount to screen price

### DIFF
--- a/contracts/vaults/BaseVaults/RibbonThetaVault.sol
+++ b/contracts/vaults/BaseVaults/RibbonThetaVault.sol
@@ -211,7 +211,10 @@ contract RibbonThetaVault is RibbonVault, RibbonThetaVaultStorage {
      * @notice Sets the new discount on premiums for options we are selling
      * @param newPremiumDiscount is the premium discount
      */
-    function setPremiumDiscount(uint256 newPremiumDiscount) external onlyKeeper {
+    function setPremiumDiscount(uint256 newPremiumDiscount)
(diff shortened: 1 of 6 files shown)


contracts/vaults/BaseVaults/RibbonThetaVault.sol
```diff
@@ -211,10 +211,13 @@ contract RibbonThetaVault is RibbonVault, RibbonThetaVaultStorage {
...
      */
-    function setPremiumDiscount(uint256 newPremiumDiscount) external onlyOwner {
+    function setPremiumDiscount(uint256 newPremiumDiscount)
+        external
+        onlyKeeper
+    {
         require(
             newPremiumDiscount > 0 &&
-                newPremiumDiscount < 100 * Vault.PREMIUM_DISCOUNT_MULTIPLIER,
+                newPremiumDiscount <= 100 * Vault.PREMIUM_DISCOUNT_MULTIPLIER,
```
kenchangh, line 220: FYI sign changed from `<` to `<=` because we want to set it to 100 (no discount)

fqlx (approved):

fqlx:
Looks good but we should add a test for setting 100% fees to test the equality operator change.

fqlx (approved):

PR #232, opened 2022-03-08
PR to fix Swap contract for Peckshield audit

stevenwal:
- PVE001: added check to avoid offer `swapId` and bid `swapId` mismatched
- PVE002: added emit SetFee event 
- PVE003: capped referrer fee at 10%


contracts/interfaces/ISwap.sol
```diff
@@ -50,6 +50,22 @@ interface ISwap {
...
 
+    struct OfferDetails {
+        // 32 byte slot 1, partial fill
+        // Seller wallet address
+        address seller;
+        // 32 byte slot 2
+        // Addess of oToken
+        address oToken;
+        // Price per oToken denominated in biddingToken
+        uint96 minPrice;
```
kenchangh, line 61: It's better to use `uint256` because EVM operates in words (32 bytes), so it saves gas

contracts/interfaces/ISwap.sol
```diff
@@ -50,6 +50,22 @@ interface ISwap {
...
+        // 32 byte slot 3
+        // ERC20 Token to bid for oToken
+        address biddingToken;
+        // Minimum oToken amount acceptable for a single bid
+        uint96 minBidSize;
```
kenchangh, line 66: Same here https://github.com/ribbon-finance/ribbon-v2/pull/232/files#r821372649

kenchangh (approved):

kenchangh (approved):

PR #233, opened 2022-03-10
feat/VaultQueue Adding VaultQueue intervault transfer

fqlx:
## feat/VaultQueue Adding VaultQueue intervault transfer

VaultQueue is a queueing system that a user initiates and our keeper finishes.  
The benefits of the VaultQueue is we simplify the user experience for transfers and withdrawing. The user only needs to queue their transfer and our keeper will finish the transfer/withdraw a week later. 

The biggest benefit is, the user doesn't need to interact with the vault a week later - one of the largest pain points in Ribbon.
  
Snowtrace:  [0x2986e874370a1398807d76d70C73E7ec7db14648](https://snowtrace.io/address/0x2986e874370a1398807d76d70C73E7ec7db14648#code)

~~Snowtrace: [0xE6325F8465dD875BE46C6997F955D6615766D8bE](https://snowtrace.io/address/0xE6325F8465dD875BE46C6997F955D6615766D8bE)~~

---

VaultQueue supports 2 transfer types.
### 1) Inter-vault transfers   
**Example:**
    - Transferring from `rETH-THETA v1` ->  `rETH-THETA v2`
    - Transferring from `rAVAX-THETA` -> `rSAVAX-THETA`

**This saves the user 2 extra steps** 
    1) initializing the withdraw
    2) waiting a week
    _3) completing the withdraw_
    _4) depositing into the new vault._
 
### 2) Withdrawals
  User queue the withdraw and the keepers competes it for them a week later.  

  **This saves the user 1 extra steps** 
    1) initializing the withdraw
    2) waiting a week
    _3) completing the withdraw_
    
Example of withdrawal AVAX to user
[0x2986e874370a1398807d76d70C73E7ec7db14648](https://snowtrace.io/tx/0x3a2639ef9df048ae56285cfb1274c6409a758de4b2f941a88a029e17359a5ef9
)
Example of withdraw with 2 users of sAVAX to users
https://snowtrace.io/tx/0xfb8c83f6244d03e32b807df6824ac5f5438925218d25115a2f6e2fca1d9a631b
    
---

### Technical Details

1. User redeems their vault shares
2. User deposits their vault shares into VaultQueue, records down how much the user is transferring
3. VaultQueue initiates a withdrawal
4. After the options position is closed at Friday 8am UTC, the VaultQueue contract completes the withdrawal for all queued withdrawals (this is triggered by the keeper).
5. The portion is disburse to each each based on the percentage of token they queued / total token.
6. When it is an inter-vault transfer:
    - The VaultQueue contract re-deposits on behalf of the user to the destination vault.
7. When it is a withdraw:
    - The VaultQueue contract deposits to the user.

Staked ETH Vault case:
- There's a special case for the staked ETH vault.  We need custom logic to support the withdraw and deposit because it's done in ETH (instead of STETH).  This means the money is deposited via the `receive() payable` function.
- We are deploying the same contract to Avax.  However the Staked ETH vault is set to the keeper address.  It's not used for sAVAX.

```
        WETH_VAULT = _wethVault; // Can be the native vault (weth or wavax)
        STETH_VAULT = _stethVault; // Must ONLY be steth vault.  (savax vault returns erc20, set as keeper)
```

**Examples:**
Withdrawal from sAVAX vault for 2 users to their wallet: https://snowtrace.io/tx/0xfb8c83f6244d03e32b807df6824ac5f5438925218d25115a2f6e2fca1d9a631b
Withdrawal from AVAX vault to single user
https://snowtrace.io/tx/0x3a2639ef9df048ae56285cfb1274c6409a758de4b2f941a88a029e17359a5ef9

---

### Planned work
- Adding a `SKIP_ROUND` feature for a user to pause for X amount of weeks (rounds) rather than withdrawing and need to deposit manually into the same vault.


contracts/utils/VaultQueue.sol
```diff
@@ -0,0 +1,118 @@
...
+        address srcVault;
+        address dstVault;
+        uint256 amount;
+        uint256 timestamp;
+    }
+}
+
+contract VaultQueue is ReentrancyGuardUpgradeable, Ownable {
+    mapping(address => IVaultQueue.Transfer[]) public qTransfer;
+    mapping(address => uint256) public totalAmount;
```
kenchangh, line 27: Do we need a `totalAmount` mapping? Because we can lookup the vault share's balance to determine how many tokens are locked into the VaultQueue

contracts/utils/VaultQueue.sol
```diff
@@ -0,0 +1,118 @@
...
+    address[] public vaults;
+
+    function interVaultTransfer() external payable onlyOwner nonReentrant {
+        for (uint256 i = 0; i < vaults.length; i++) {
+            IRibbonVault(vaults[i]).completeWithdraw();
```
kenchangh, line 32: Since it's likely that not *all* vaults will have withdrawals, this needs to check if there is an open withdrawal for the vault before calling `completeWithdraw`

contracts/utils/VaultQueue.sol
```diff
@@ -0,0 +1,118 @@
...
+        }
+    }
+
+    function queueTransfer(
+        address srcVault,
+        address dstVault,
+        uint256 amount
+    ) external nonReentrant {
+        for (uint256 i = 0; i < vaults.length; i++) {
+            for (uint256 j = 0; j < qTransfer[vaults[i]].length; j++) {
```
kenchangh, line 63: Worth refactoring this logic out to a view so the frontend can check too

contracts/utils/VaultQueue.sol
```diff
@@ -0,0 +1,118 @@
...
+        address creditor;
+        address srcVault;
+        address dstVault;
+        uint256 amount;
+        uint256 timestamp;
+    }
+}
+
+contract VaultQueue is ReentrancyGuardUpgradeable, Ownable {
+    mapping(address => IVaultQueue.Transfer[]) public qTransfer;
```
kenchangh, line 26: I think it may make sense to fix the Transfer[] array size to a number so we can bound the number of transfers.

contracts/utils/VaultQueue.sol
fqlx: Done

contracts/utils/VaultQueue.sol
fqlx: Instead of fixing the array size, we could limit the queue array by checking `require(qTransfer.length < 256);` in the queuing function.

contracts/utils/VaultQueue.sol
```diff
@@ -0,0 +1,118 @@
...
+    struct Transfer {
+        address creditor;
+        address srcVault;
+        address dstVault;
+        uint256 amount;
+        uint256 timestamp;
+    }
+}
+
+contract VaultQueue is ReentrancyGuardUpgradeable, Ownable {
```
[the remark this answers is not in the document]
fqlx: Done

contracts/utils/VaultQueue.sol
fqlx: This is checked at the beginning of `_completeWithdraw()` and it seems like it adds more gas to check, since it would be checked again regardless.

fqlx: Done

contracts/utils/VaultQueue.sol
```diff
@@ -0,0 +1,229 @@
...
+        address srcVault;
+        address dstVault;
+        address depositContract;
+        TransferAction action;
+        uint256 amount;
+        uint256 timestamp;
+    }
+}
+
+contract VaultQueue is
```
0xTomoyo, line 36: I think you can inherit from `IVaultQueue` to avoid having to use `IVaultQueue.Transfer` (can just use `Transfer` directly instead)

contracts/utils/VaultQueue.sol
```diff
@@ -0,0 +1,229 @@
...
+    ReentrancyGuardUpgradeable
+{
+    using SafeERC20 for IERC20;
+
+    mapping(address => IVaultQueue.Transfer[]) public qTransfer;
+    mapping(address => uint256) public totalAmount;
+    address[] public vaults;
+
+    address public wethVault = 0x25751853Eab4D0eB3652B5eB6ecB102A2789644B;
+    address public stEthVault = 0x53773E034d9784153471813dacAFF53dBBB78E8c;
```
0xTomoyo, line 48: Probably should pass these 2 vars in from a constructor instead of hardcoding them

contracts/utils/VaultQueue.sol
```diff
@@ -0,0 +1,229 @@
...
+
+    // ETH and ERC20s are not kept in this contract unless explicitly sent.
+    // This contract intermediately holds tokens only during transfer() and only within the transaction.
+    function rescueETH(uint256 amount) external onlyOwner {
+        payable(msg.sender).transfer(amount);
+    }
+
+    function rescue(address asset, uint256 amount) external onlyOwner {
+        IERC20(asset).approve(address(this), amount);
+        IERC20(asset).transferFrom(address(this), msg.sender, amount);
```
0xTomoyo, line 225: Can be replaced with ```IERC20(asset).transfer(msg.sender, amount);```

contracts/utils/VaultQueue.sol
fqlx: I was considering this but the vault behavior on other chains is different.  e.g. AVAX and sAVAX.

contracts/utils/VaultQueue.sol
fqlx: done

contracts/utils/VaultQueue.sol
fqlx: done

contracts/utils/VaultQueue.sol
fqlx: done

contracts/utils/VaultQueue.sol
```diff
@@ -0,0 +1,228 @@
...
+    Initializable,
+    OwnableUpgradeable,
+    ReentrancyGuardUpgradeable
+{
+    enum TransferAction {INTERVAULT, WITHDRAW}
+    struct Transfer {
+        address creditor;
+        address srcVault;
+        address dstVault;
+        address depositContract;
```
kenchangh, line 33: Maybe we can completely remove the need to store the `depositContract` by hardcoding the mapping of `dstVault` to the DepositHelper contract. Since this edge case only exists for savax currently

contracts/utils/VaultQueue.sol
```diff
@@ -0,0 +1,228 @@
...
+        TransferAction action;
+        uint256 amount;
+        uint256 timestamp;
```
kenchangh, line 36: `timestamp` could be stored as a uint32

contracts/utils/VaultQueue.sol
```diff
@@ -0,0 +1,228 @@
...
+import {Vault} from "../libraries/Vault.sol";
+import {DSMath} from "../vendor/DSMath.sol";
+
+contract VaultQueue is
+    Initializable,
+    OwnableUpgradeable,
+    ReentrancyGuardUpgradeable
+{
+    enum TransferAction {INTERVAULT, WITHDRAW}
+    struct Transfer {
```
kenchangh, line 29: To optimize this more, we can fit the Struct into 3 32-byte slots with [struct packing](https://fravoll.github.io/solidity-patterns/tight_variable_packing.html) 

```solidity
struct Transfer {
  // slot 1
  address creditor;
  // slot 2
   address dstVault;
  TransferAction action;
  uint32 timestamp;
  // slot 3
  uint256 amount;
}
```

contracts/utils/VaultQueue.sol
```diff
@@ -0,0 +1,228 @@
...
+        address _stethEthCrvPool
+    ) external initializer {
+        __ReentrancyGuard_init();
+        __Ownable_init();
+
+        transferOwnership(msg.sender);
+
+        wethVault = _wethVault; // Can be the native vault (weth or wavax)
+        stethVault = _stethVault; // Must ONLY be steth vault.  (savax vault returns erc20, set as keeper)
+        stethEthCrvPool = _stethEthCrvPool;
```
kenchangh, line 62: We're not using the `stethEthCrvPool` as part of the withdrawal so we can remove this entirely

contracts/utils/VaultQueue.sol
fqlx: Sure but keep in mind this would use more gas.  I'll make the change because I prefer readability over gas optimizations right now

contracts/utils/VaultQueue.sol
```diff
@@ -0,0 +1,223 @@
...
+import {ICRV} from "../interfaces/ICRV.sol";
+import {Vault} from "../libraries/Vault.sol";
+import {DSMath} from "../vendor/DSMath.sol";
+
+contract VaultQueue is
+    Initializable,
+    OwnableUpgradeable,
+    ReentrancyGuardUpgradeable
+{
+    enum TransferAction {INTERVAULT, WITHDRAW}
```
kenchangh, line 28: Thinking about it, why do we need `TransferAction`? Because if the `srcVault` and `dstVault` are the same, that implies that it is an `INTERVAULT` transfer, so it feels redundant

contracts/utils/VaultQueue.sol
```diff
@@ -0,0 +1,223 @@
...
+    function withdrawToCreditor(
+        address vault,
+        address creditor,
+        uint256 portion
+    ) private {
+        if (vault == stethVault) {
+            address steth = IRibbonVault(vault).STETH();
+            IERC20(steth).transfer(creditor, portion);
+        } else if (vault == wethVault) {
+            payable(creditor).transfer(portion);
```
kenchangh, line 91: The better way to send ether is to use `payable(creditor).call{value: portion}("")`

contracts/utils/VaultQueue.sol
```diff
@@ -0,0 +1,223 @@
...
+
+    function reset(address vault) external onlyOwner {
+        delete qTransfer[vault];
+        totalAmount[vault] = 0;
+    }
+
+    // ETH and ERC20s are not kept in this contract unless explicitly sent.
+    // This contract intermediately holds tokens only during transfer() and only within the transaction.
+    function rescueETH(uint256 amount) external onlyOwner {
+        payable(msg.sender).transfer(amount);
```
kenchangh, line 214: Same here https://github.com/ribbon-finance/ribbon-v2/pull/233/files#r843423955

contracts/utils/VaultQueue.sol
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    function rescue(address asset, uint256 amount) external onlyOwner {
+        IERC20(asset).approve(address(this), amount);
```
kenchangh, line 218: We don't have to call `approve` before transferring

contracts/utils/VaultQueue.sol
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    function reset(address vault) external onlyOwner {
+        delete qTransfer[vault];
+        totalAmount[vault] = 0;
+    }
+
+    // ETH and ERC20s are not kept in this contract unless explicitly sent.
+    // This contract intermediately holds tokens only during transfer() and only within the transaction.
+    function rescueETH(uint256 amount) external onlyOwner {
```
kenchangh, line 213: TBH I don't think we should have a rescue function because this gives us the capability to rugpull the contract.

kenchangh: Since it is upgradeable, if there is major contract bug, we can recover from it by upgrading with a fix.

contracts/utils/VaultQueue.sol
```diff
@@ -0,0 +1,223 @@
...
+        );
+        totalAmount[srcVault] += amount;
+    }
+
+    function hasWithdrawal(address vault, address user)
+        public
+        view
+        returns (bool)
+    {
+        for (uint256 j = 0; j < qTransfer[vault].length; j++) {
```
kenchangh, line 192: To optimize this loop, you can do:

```solidity
uint256 qTransferLen = qTransfer[vault].length;
for (uint256 j = 0; j < qTransferLen; ++j) {
```

contracts/utils/VaultQueue.sol
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    function disburse(address vault) private {
+        uint256 balance = getInterVaultBalance(vault);
+        uint256 len = qTransfer[vault].length;
+        uint256 totalAmt = totalAmount[vault];
+        totalAmount[vault] = 0;
+
+        for (uint256 j = len; j > 0; j--) {
+            Transfer memory queue = pop(qTransfer[vault]);
```
kenchangh, line 114: Btw it's cheaper if you operated on `storage` instead of `memory` so it's cheaper if you did `Transfer storage queue`

contracts/interfaces/IRibbon.sol
```diff
@@ -9,9 +9,34 @@ interface IRibbonVault {
 
     function cap() external view returns (uint256);
 
+    function WETH() external view returns (address);
+
+    function withdrawals(address) external view returns (uint256);
```
kenchangh, line 14: Btw this function signature is wrong, it should be 

```solidity
struct Withdrawal {
        uint16 round;
        uint128 shares;
}

function withdrawals(address) external view returns (Withdrawal);
```

contracts/utils/VaultQueue.sol
```diff
@@ -0,0 +1,223 @@
...
+        address dstVault,
+        address depositContract,
+        TransferAction transferAction,
+        uint256 amount
+    ) external nonReentrant {
+        require(
+            !hasWithdrawal(srcVault, msg.sender),
+            "Withdraw already submitted"
+        );
+        require(qTransfer[srcVault].length < 256, "Transfer queue full");
```
kenchangh, line 157: Is 256 too high of a limit? Think it should be like 20 or something. Also we should set this as a constant so it's obvious as a limit

contracts/utils/VaultQueue.sol
```diff
@@ -0,0 +1,223 @@
...
+
+        if (transferAction == TransferAction.INTERVAULT) {
+            require(depositContract != address(0), "Intervault !address");
```
kenchangh, line 160: Nit, don't think this revert message is helpful, maybe something like `No deposit contract`

contracts/utils/VaultQueue.sol
```diff
@@ -0,0 +1,223 @@
...
+        return false;
+    }
+
+    function pop(Transfer[] storage array) private returns (Transfer memory) {
+        Transfer memory item = array[array.length - 1];
+        array.pop();
+        return item;
+    }
+
+    function reset(address vault) external onlyOwner {
```
kenchangh, line 206: Just thinking about this function more, I think this function could be potentially destructive so I'm in favor of removing it

contracts/utils/VaultQueue.sol
```diff
@@ -0,0 +1,223 @@
...
+    function rescueETH(uint256 amount) external onlyOwner {
+        payable(msg.sender).transfer(amount);
+    }
+
+    function rescue(address asset, uint256 amount) external onlyOwner {
+        IERC20(asset).approve(address(this), amount);
+        IERC20(asset).transfer(msg.sender, amount);
+    }
+
+    receive() external payable {}
```
kenchangh, line 222: Looping back to what we talked about before, let's add a check to verify that it's coming from the vault

```solidity
require(msg.sender == wethVault || msg.sender == stethVault);
```

contracts/utils/VaultQueue.sol
fqlx: I was planning to add `SKIP_ROUND` and different validation checks for each `TransferAction`

contracts/utils/VaultQueue.sol
fqlx: This was incase someone sent ETH to the contract and we needed to recover it, instead I will a validation check in the `payable` function.

contracts/interfaces/IRibbon.sol
fqlx: Good catch!

contracts/utils/VaultQueue.sol
fqlx: Removed `rescue()`

contracts/utils/VaultQueue.sol
fqlx: Removed `rescueEth()`

contracts/utils/VaultQueue.sol
fqlx: done

contracts/utils/VaultQueue.sol
fqlx: done

contracts/utils/VaultQueue.sol
fqlx: done

contracts/utils/VaultQueue.sol
fqlx: Default is 32 and we can adjust it by calling `setQueueSize()`

contracts/utils/VaultQueue.sol
```diff
@@ -0,0 +1,228 @@
...
+
+contract VaultQueue is
+    Initializable,
+    OwnableUpgradeable,
+    ReentrancyGuardUpgradeable
+{
+    enum TransferAction {INTERVAULT, WITHDRAW}
+    struct Transfer {
+        address creditor;
+        address srcVault;
```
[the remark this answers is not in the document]
fqlx: `srcVault` is no longer needed.  I refactored the code to pass the `srcVault` and didn't remove it from the `Transfer` struct.

contracts/utils/VaultQueue.sol
fqlx: I think the upside of leaving it here to support future contracts outweighs the minimal gas we'll spend removing it

contracts/utils/VaultQueue.sol
fqlx: This doesn't seem possible with how the pop mutates the transfer queue state

contracts/utils/VaultQueue.sol
```diff
@@ -43,6 +42,7 @@ contract VaultQueue is
 
     address public wethVault;
     address public stethVault;
+    uint256 public queueSize;
```
kenchangh, line 45: these need to be immutable variables

contracts/utils/VaultQueue.sol
```diff
@@ -43,6 +42,7 @@ contract VaultQueue is
 
     address public wethVault;
     address public stethVault;
```
kenchangh, line 44: These need to be immutable variables that are set on constructor

fqlx: You can't use immutable in `initialize()` and can't use their constructors with upgradeable contracts.

https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable

>You can use your Solidity contracts with OpenZeppelin Upgrades without any modifications, except for their constructors

contracts/utils/VaultQueue.sol
fqlx: Removed dstVault

kenchangh:
Stale

PR #235, opened 2022-03-19
feat/APE-THETA Ape Call Vault

fqlx:
- feat/APE-THETA Ape Call Vault


scripts/utils/constants.ts
```diff
@@ -8,6 +8,7 @@ export const SAVAX_STRIKE_STEP = 10;
 export const AAVE_STRIKE_STEP = 10;
 export const NEAR_STRIKE_STEP = 5;
 export const AURORA_STRIKE_STEP = 5;
+export const APE_STRIKE_STEP = 5;
```
kenchangh, line 11: This should be 1 for APE

fqlx: fixed

PR #236, opened 2022-04-03
task/rm-aurora-pipeline Removing Aurora/Near from the tests to speedup the pipeline

fqlx:
task/rm-aurora-pipeline Removing Aurora/Near from the tests to speedup the pipeline

    - Left the core code, may delete in the future
    - Also ran prettier


stevenwal (approved):
LGTM

PR #239, opened 2022-04-06
Update the premium to look up on auction start

kenchangh:
- `startAuction` does not restart a new auction with an updated pricing, but is stuck with an old pricing
- This makes it so that it looks up the most updated price from the OptionsPremiumPricer

### DIFF
--- a/contracts/libraries/VaultLifecycleSTETH.sol
+++ b/contracts/libraries/VaultLifecycleSTETH.sol
@@ -77,13 +77,12 @@ library VaultLifecycleSTETH {
             false
         );
 
-        uint256 premium =
-            _getOTokenPremium(
-                otokenAddress,
-                closeParams.optionsPremiumPricer,
-                closeParams.premiumDiscount,
-                collateralAsset
-            );
+        premium = _getOTokenPremium(
+            otokenAddress,
+            closeParams.optionsPremiumPricer,
+            closeParams.premiumDiscount,
+            collateralAsset
+        );
 
         return (otokenAddress, premium, strikePrice, delta);
     }
--- a/contracts/libraries/VaultLifecycleSTETH.sol
+++ b/contracts/libraries/VaultLifecycleSTETH.sol
@@ -21,6 +21,7 @@ import {
     GammaTypes
 } from "../interfaces/GammaInterface.sol";
 import {IERC20Detailed} from "../interfaces/IERC20Detailed.sol";
+import {IOptionsPremiumPricer} from "../interfaces/IRibbon.sol";
 
 library VaultLifecycleSTETH {
     using SafeMath for uint256;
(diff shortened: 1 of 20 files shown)


test/RibbonThetaVault.ts
```diff
@@ -3138,6 +3138,112 @@ function behavesLikeRibbonOptionsVault(params: {
...
+          .setPremiumDiscount(BigNumber.from("800"));
+
+        await vault.connect(keeperSigner).startAuction();
+
+        assert.bnEqual(
+          await otoken.balanceOf(gnosisAuction.address),
+          initialOtokenBalance
+        );
+
+        // otoken price is decreased on the auction
```
chuddster, line 3193: is it possible it is less than bc price of underlying changed (so relative to same strike it could be cheaper). maybe make sure it is less than by the premium discount at least

contracts/vaults/BaseVaults/RibbonThetaVault.sol
```diff
@@ -428,11 +428,16 @@ contract RibbonThetaVault is RibbonVault, RibbonThetaVaultStorage {
     function _startAuction() private {
         GnosisAuction.AuctionDetails memory auctionDetails;
 
-        uint256 currOtokenPremium = currentOtokenPremium;
-
+        address currentOtoken = optionState.currentOption;
+        uint256 currOtokenPremium =
```
chuddster, line 432: might make sense to remove redundant calculation in vaultlifecycle. it will probably make up for contract redeployments quickly as its an expensive calculation.

contracts/vaults/BaseVaults/RibbonThetaVault.sol
```diff
@@ -428,11 +428,16 @@ contract RibbonThetaVault is RibbonVault, RibbonThetaVaultStorage {
     function _startAuction() private {
```
fqlx, line 428: We need to make this change in the other vault contracts too.
`STETHVault`, `TreasuryVault`, `YearnVaults`

contracts/vaults/BaseVaults/RibbonThetaVault.sol
fqlx: Agreed

test/RibbonThetaVault.ts
kenchangh: Don't really want to add another restriction so we have more flexibility here, for whatever reasons the auction doesn't fill. eg maybe there just aren't bidders so we need to readjust the price higher or lower.

contracts/libraries/VaultLifecycleSTETH.sol
```diff
@@ -569,4 +558,61 @@ library VaultLifecycleSTETH {
...
+        return
+            _getOTokenPremium(
+                oTokenAddress,
+                optionsPremiumPricer,
+                premiumDiscount,
+                collateralToken
+            );
+    }
+
+    function _getOTokenPremium(
```
kenchangh, line 577: We had to duplicate this function across all the VaultLifecycle's because we have to link the 2 libraries otherwise. On top of that, we also do the extra calculation of multiplying the pricePerShare of the collateral token

stevenwal (dismissed):
LGTM

stevenwal (approved):

PR #240, opened 2022-04-11
Fix issue where 0 premium is distributed to withdrawing user

stevenwal:
This is to fix 3 main issues with the current Ribbon Treasury Vault

1. We noticed an issue where the premium is not completely distributed as withdrawing users' shares have not been burned during distribution. 
(Example Tx: https://etherscan.io/tx/0x365d120375f50d6a67a938ef157a2a3968b48e35ae16b5c76cfd36612d6b7731). 
Solution: 
    - reduce total supply by last queued withdraw shares
    - remove depositors during initiateWithdraw if they have 0 leftover shares

2. During distribution, dust amount might be left in the vault. This would trigger distribution of close to 0 amounts to depositors.
(Example Tx: https://etherscan.io/tx/0xd49ba732b1d12da87d33b31edc3f34ba0115f28612d76cf6b639bf63827ad88e)
Solution: 
    - add a minimum amount of 10 USDC before distribution can be called

3. In the event someone transfer all their oTokens to someone else, their address will sit in the depositors list permanently.
Solution:
     - disable transfer of rToken for treasury vault

### DIFF
--- a/contracts/vaults/TreasuryVault/RibbonTreasuryVault.sol
+++ b/contracts/vaults/TreasuryVault/RibbonTreasuryVault.sol
@@ -75,6 +75,9 @@ contract RibbonTreasuryVault is
     // The minimum duration for an option auction.
     uint256 private constant MIN_AUCTION_DURATION = 5 minutes;
 
+    // The minimum amount above which premium distribution will occur during commitAndClose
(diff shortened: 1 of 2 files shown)


kenchangh (approved):
LGTM

PR #241, opened 2022-04-14
OptionsPurchaseQueue contract

0xTomoyo:
## Gnosis Easy Auction settlement price calculation

1. Get the `clearingPriceOrder` for the auction
2. Decode the order and get the `sellAmount` and `sellAmount`
3. `settlemenPrice = (10**8) * sellAmount / sellAmount` (since oTokens have 8 decimals)

### `requestPurchase`
Allows anyone to request to purchase options from a Theta vault (before the options are minted.

### `allocateOptions`
The Theta vault calls this after it mints the options. This transfers the requested options over to the OptionsPurchaseQueue.

### `sellToBuyers`
The Theta vault calls this after it auctions off the options it didn't transfer to the queue. It passes in the price the auction settled at, and processes the token transfers between the the Theta vault and the buyers.

### `cancelAllPurchases`
Allows the owner to cancel all requested purchases for a vault and refund users.


contracts/utils/OptionsPurchaseQueue.sol
```diff
@@ -0,0 +1,442 @@
...
+     * @param vault The vault to purchase options from
+     * @param optionsAmount Amount of options requested
+     * @return premiums Amount of premiums transferred from the buyer
+     */
+    function requestPurchase(address vault, uint256 optionsAmount)
+        external
+        override
+        returns (uint256)
+    {
+        require(whitelistedBuyer[msg.sender], "Buyer not whitelisted");
```
fqlx, line 149: Probably better to set an array limit on `purchases[vault]` than whitelist the buyer so we don't have an unbounded for loop

contracts/vaults/BaseVaults/RibbonThetaVault.sol
```diff
@@ -408,12 +411,23 @@ contract RibbonThetaVault is RibbonVault, RibbonThetaVaultStorage {
...
+                GAMMA_CONTROLLER,
+                MARGIN_POOL,
+                newOption,
+                lockedBalance
+            );
+
+        address _optionsPurchaseQueue = optionsPurchaseQueue;
+        if (_optionsPurchaseQueue != address(0)) {
+            // Allocate a maximum of 50% of options towards the purchase queue
+            uint256 allocatedOptions = optionsMintAmount.div(2);
```
fqlx, line 425: Write a setter for what % we allocate the options.  I think 50% is too much and better if we can change it in the future than fix it to 10%, 20%, 50%, etc.. Not sure how popular this will be

contracts/utils/OptionsPurchaseQueue.sol
```diff
@@ -0,0 +1,442 @@
...
+
+    /************************************************
+     *  STORAGE
+     ***********************************************/
+
+    /// @notice Stores the purchase queue for each vault
+    mapping(address => Purchase[]) public purchases;
+
+    /// @notice Stores the total options being purchased for each vault
+    mapping(address => uint256) public override totalOptionsAmount;
```
kenchangh, line 46: Is this `otoken address => otoken amount`?

contracts/utils/OptionsPurchaseQueue.sol
```diff
@@ -0,0 +1,442 @@
...
+
+    /// @notice Stores the amount of options the vault is allocating towards the purchase queue
+    /// @dev When vaultAllocatedOptions != 0, new purchase requests for the vault are blocked since the vault has
```
kenchangh, line 49: nit, do you mean `When vaultAllocatedOptions == 0`?

contracts/utils/OptionsPurchaseQueue.sol
```diff
@@ -0,0 +1,442 @@
...
+        require(whitelistedBuyer[msg.sender], "Buyer not whitelisted");
+        uint256 _ceilingPrice = ceilingPrice[vault];
+        require(_ceilingPrice != 0, "Invalid vault");
+        // This prevents new purchase requested after the vault has set its allocation
+        require(vaultAllocatedOptions[msg.sender] == 0, "Vault allocated");
+        require(optionsAmount != 0, "!optionsAmount");
+
+        // premiums = optionsAmount * ceilingPrice
+        uint256 premiums =
+            optionsAmount.mul(_ceilingPrice).div(10**Vault.OTOKEN_DECIMALS);
```
kenchangh, line 158: We can remove all instances of SafeMath because of Solidity 0.8 :)

kenchangh, line 158: Also, shouldn't this be divided by the asset's decimal? Like:

```solidity
uint premiums = (optionsAmount * ceilingPrice) / assetDecimals
```

So for buying 1 put option with ceiling price of 15 USDC, it would be:

(1*10**8 * 15*10**6) / (10**6)

contracts/utils/OptionsPurchaseQueue.sol
```diff
@@ -0,0 +1,442 @@
...
+        // Blocks new purchase requests until sellToBuyers() is called
+        vaultAllocatedOptions[msg.sender] = allocatedOptions;
+
+        if (allocatedOptions != 0) {
+            // Transfer allocated options from the vault to this contract
+            IERC20(IRibbonThetaVault(msg.sender).currentOption())
+                .safeTransferFrom(msg.sender, address(this), allocatedOptions);
+        }
+
+        emit OptionsAllocated(msg.sender, allocatedOptions);
```
kenchangh, line 216: Should the emit should go under the `if` branch?

contracts/vaults/BaseVaults/RibbonThetaVault.sol
0xTomoyo: It only allocates up to 50% (see the `allocateOption()` function

Example 1:
Lets say the vault mints 1000 options. On the purchase queue there are a total of 300 options requested for purchase. Only 300 will get allocated to the purchase queue.

Example 2:
Lets say the vault mints 1000 options. On the purchase queue there are a total of 700 options requested for purchase. Only 500 will get allocated to the purchase queue.

contracts/utils/OptionsPurchaseQueue.sol
0xTomoyo: I think we can put it put it under since there may be a situation where the vault allocates 0 options to the queue

contracts/utils/OptionsPurchaseQueue.sol
0xTomoyo: Since the ceilingPrice is denominated in the asset's decimals already, it should be divided by the oToken's decimal

Example: Buying 10 put options with ceiling price of 15 USDC
`10e8 * 15e6 / 1e8 = 150e6 (150000000)`

0xTomoyo: lol true

contracts/utils/OptionsPurchaseQueue.sol
0xTomoyo: Vault address => otoken amount 

(since we don't know the oToken address until Friday)

This is reset after the options get sold to queue.

contracts/utils/OptionsPurchaseQueue.sol
0xTomoyo: New purchase requests get blocked when `vaultAllocatedOptions != 0`, see: https://github.com/ribbon-finance/ribbon-v2/pull/241/files#diff-6d059eef33889a02d96f146813863a0b1a958de6e3730d8ef14ecdc88f4db0e3R144-R145

contracts/utils/OptionsPurchaseQueue.sol
0xTomoyo: Switched to using a min purchase requirement for each purchase request

contracts/utils/OptionsPurchaseQueue.sol
```diff
@@ -0,0 +1,660 @@
...
+
+        emit BuyerWhitelisted(buyer);
+    }
+
+    /**
+     * @notice Removes a buyer from the purchase queue whitelist
+     * @dev Only callable by the owner
+     * @param buyer The buyer to blacklist
+     */
+    function blacklistBuyer(address buyer) external override onlyOwner {
```
kargakis, line 561: Remove from whitelist does not necessarily mean blacklisting. This is slightly confusing since the buyer can still place purchase requests that meet the minimum requirement.

contracts/libraries/VaultLifecycle.sol
```diff
@@ -716,6 +720,94 @@ library VaultLifecycle {
...
+    ) external returns (uint256 allocatedOptions) {
+        // Skip if optionsPurchaseQueue is address(0)
+        if (optionsPurchaseQueue != address(0)) {
+            allocatedOptions = optionsAmount.mul(optionAllocation).div(
+                100 * Vault.OPTION_ALLOCATION_MULTIPLIER
+            );
+            allocatedOptions = IOptionsPurchaseQueue(optionsPurchaseQueue)
+                .getOptionsAllocation(address(this), allocatedOptions);
+
+            if (allocatedOptions != 0) {
```
stevenwal, line 747: Do we need to have this check? We have a similar check here: https://github.com/ribbon-finance/ribbon-v2/pull/241/files#diff-6d059eef33889a02d96f146813863a0b1a958de6e3730d8ef14ecdc88f4db0e3R207-R216

stevenwal, line 747: Do we need to have this check? We have a similar check here: https://github.com/ribbon-finance/ribbon-v2/pull/241/files#diff-6d059eef33889a02d96f146813863a0b1a958de6e3730d8ef14ecdc88f4db0e3R207-R216

0xTomoyo: yeah, it avoids the vault calling the purchase queue altogether

contracts/storage/RibbonDeltaVaultStorage.sol
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: MIT
+pragma solidity =0.8.4;
+import {IRibbonThetaVault} from "../interfaces/IRibbonThetaVault.sol";
+import {Vault} from "../libraries/Vault.sol";
+
+abstract contract RibbonDeltaVaultStorageV1 {
```
kenchangh, line 6: Should this not be RibbonGammaVaultStorage instead?

kenchangh (approved):

PR #247, opened 2022-04-27
Upgrade changes - recoverTokens, remove minETHOut

kenchangh:


contracts/vaults/BaseVaults/RibbonThetaVault.sol
```diff
@@ -485,4 +485,21 @@ contract RibbonThetaVault is RibbonVault, RibbonThetaVaultStorage {
             uint256(vaultState.lockedAmount).sub(unlockedAssetAmount)
         );
     }
+
+    /**
+     * @notice Recovery function that returns an ERC20 token to the recipient
+     */
+    function recoverTokens(address token, address recipient)
```
stevenwal, line 494: Should we add quantity as a param here? In case there are 2 people depositing the wrong assets. 

Otherwise, we can also recover to our account and re-distribute to the users.

test/RibbonThetaSTETHVault.ts
```diff
@@ -2723,8 +2723,6 @@ function behavesLikeRibbonOptionsVault(params: {
     });
 
     describe("#completeWithdraw", () => {
```
stevenwal, line 2725: Can we update the same on upgrade test?

stevenwal (dismissed):

stevenwal (approved):

PR #254, opened 2022-05-21
[WIP] Vault Pauser

StephKua:


contracts/vaults/VaultPauser/RibbonVaultPauser.sol
```diff
@@ -0,0 +1,253 @@
...
+
+        currentVault.depositFor(totalWithdrawAmount, currentUser);
+    }
+
+    /**
+     * @notice process withdrawals by completing in a batch
+     * @param _vaultAddress vault's address to be processed
+     */
+    function processWithdrawal(address _vaultAddress) external onlyKeeper {
+        _processWithdrawal(_vaultAddress);
```
kenchangh, line 240: We can move the entire function body within the `processWithdrawal` instead, since `_processWithdrawal` is not used anywhere else

contracts/vaults/VaultPauser/RibbonVaultPauser.sol
```diff
@@ -0,0 +1,253 @@
...
+        STETH = _steth;
+    }
+
+    // /**
+    //  * @notice Initializes the contract with storage variables.
+    //  */
+    function initialize(address _owner, address _keeper) external initializer {
+        __ReentrancyGuard_init();
+        __Ownable_init();
+        transferOwnership(_owner);
```
kenchangh, line 97: We don't need to use the proxy contract pattern for the VaultPauser contract, can just use a normal constructor

contracts/vaults/VaultPauser/RibbonVaultPauser.sol
```diff
@@ -0,0 +1,253 @@
...
+
+    /// @notice Stores all the vault's paused positions
+    struct PauseReceipt {
+        uint16 round;
+        address account;
+        uint128 shares;
+    }
+
+    mapping(address => mapping(address => PauseReceipt[]))
+        public pausedPositions;
```
kenchangh, line 49: For simplicity, could we just do `address => PauseReceipt`? So we make it so that one user can only have one pause per vault. ie they can only pause their entire balance

contracts/vaults/BaseVaultWithSwap/RibbonThetaVaultWithSwap.sol
```diff
@@ -501,4 +510,17 @@ contract RibbonThetaVaultWithSwap is RibbonVault, RibbonThetaVaultStorage {
...
+    /**
+     * @notice pause a user's vault position
+     */
+    function pausePosition() external {
+        address _vaultPauser = vaultPauser;
+        require(_vaultPauser != address(0)); // Removed revert msgs due to contract size limit
+        _redeem(0, true);
+        uint256 heldByAccount = balanceOf(msg.sender);
+        _approve(msg.sender, _vaultPauser, heldByAccount);
+        _transfer(msg.sender, _vaultPauser, heldByAccount);
```
kenchangh, line 523: We shouldn't be transferring the vault shares from the vault itself, this has to be done within the VaultPauser contract with `transferFrom`

contracts/vaults/VaultPauser/RibbonVaultPauser.sol
```diff
@@ -0,0 +1,253 @@
...
+     * @notice pause position from vault by redeem all the shares from vault to Pauser
+     * @param _account user's address
+     * @param _amount the amount of shares
+     */
+    function pausePosition(address _account, uint256 _amount)
+        external
+        override
+    {
+        address currentVaultAddress = address(msg.sender);
+        IRibbonThetaVault currentVault = IRibbonThetaVault(currentVaultAddress);
```
kenchangh, line 151: Do a `IERC20(currentVaultAddress).transferFrom(_account, _amount)` here

contracts/vaults/VaultPauser/RibbonVaultPauser.sol
```diff
@@ -0,0 +1,253 @@
...
+        delete pausedPositions[_vaultAddress][currentUser];
+
+        string memory currentSymbol = currentVault.symbol();
+
+        // stETH transfers suffer from an off-by-1 error
+        // since we received STETH , we shall deposit using STETH instead of ETH
+        if (
+            (keccak256(abi.encodePacked(currentSymbol)) ==
+                keccak256(abi.encodePacked("rSTETH-THETA")))
+        ) {
```
kenchangh, line 213: We can avoid comparing strings by just hardcoding the stETH vault address itself

```solidity
address constant STETH_VAULT = 0x53773E034d9784153471813dacAFF53dBBB78E8c

if currentVault == STETH_VAULT {
}
```

contracts/vaults/VaultPauser/RibbonVaultPauser.sol
```diff
@@ -0,0 +1,253 @@
...
+            currentVault.depositYieldToken(totalWithdrawAmount, currentUser);
+            return;
+        }
+
+        emit Resume(currentUser, _vaultAddress, totalWithdrawAmount);
+        // if asset is ETH, we will convert it into WETH before depositing
+        if (currentVault.vaultParams().asset == WETH) {
+            IWETH(WETH).deposit{value: totalWithdrawAmount}();
+        }
+        IERC20(currentVault.vaultParams().asset).approve(
```
kenchangh, line 227: Each call to `currentVault.vaultParams().asset` costs a lot of gas since they access storage variables. You will have to cache them into a variable

```solidity
address asset = currentVault.vaultParams().asset
```

contracts/vaults/VaultPauser/RibbonVaultPauser.sol
```diff
@@ -0,0 +1,253 @@
...
+        IRibbonThetaVault currentVault = IRibbonThetaVault(currentVaultAddress);
+
+        PauseReceipt[] storage pauseReceipts =
+            pausedPositions[currentVaultAddress][_account];
+
+        // user can pause multiple position hence it's necessary to make an array
+        // to store each paused position
+        pauseReceipts.push(
+            PauseReceipt({
+                round: uint16(currentVault.vaultState().round),
```
kenchangh, line 160: Same for `currentVault.vaultState().round`

https://github.com/ribbon-finance/ribbon-v2/pull/254/files#r884435142

contracts/vaults/VaultPauser/RibbonVaultPauser.sol
```diff
@@ -150,26 +141,30 @@ contract RibbonVaultPauser is
...
-            currentVault.vaultState().round
+            address(this),
+            _amount
         );
 
+        uint16 round = currentVault.vaultState().round;
+
+        pausedPositions[currentVaultAddress][_account] = PauseReceipt({
+            round: round,
+            account: address(_account),
```
kenchangh, line 161: We can remove `account` from the PauseReceipt because we're already using `account` to identify the PauseReceipt

contracts/vaults/VaultPauser/RibbonVaultPauser.sol
```diff
@@ -150,26 +141,30 @@ contract RibbonVaultPauser is
...
+            shares: uint104(_amount),
+            paused: true
```
kenchangh, line 163: `paused` doesn't need to exist either because if the `PauseReceipt` is not null, it is paused.

kenchangh: We can delete it from the user's pausedPositions once they call resumePosition

contracts/vaults/STETHVault/base/RibbonVault.sol
```diff
@@ -331,7 +331,24 @@ contract RibbonVault is
...
+
+        IERC20(STETH).safeTransferFrom(msg.sender, address(this), amount);
+    }
+
+    /**
+     * @notice Deposits the `collateralAsset` into the contract and mint vault shares.
+     * @param amount is the amount of `collateralAsset` to deposit
+     * @param creditor is the amount of `collateralAsset` to deposit
+     */
+    function depositYieldTokenFor(uint256 amount, address creditor)
```
kenchangh, line 348: Had to add this function because the dapps rely on `depositYieldToken`

contracts/vaults/VaultPauser/RibbonVaultPauser.sol
```diff
@@ -179,31 +180,42 @@ contract RibbonVaultPauser is Ownable, IVaultPauser {
...
                 currentParams.decimals
             );
 
+        // delete position once transfer (revert to zero)
+        delete pausedPositions[_vaultAddress][currentUser];
+
         // stETH transfers suffer from an off-by-1 error
         // since we received STETH , we shall deposit using STETH instead of ETH
         if (_vaultAddress == STETH_VAULT) {
             totalWithdrawAmount = totalWithdrawAmount.sub((3));
```
kenchangh: If possible, I'd like us to avoid -3

contracts/vaults/VaultPauser/RibbonVaultPauser.sol
```diff
@@ -0,0 +1,249 @@
...
+     * @notice resume user's position into vault by making a deposit
+     * @param _vaultAddress vault's address
+     */
+    function resumePosition(address _vaultAddress) external override {
+        IRibbonThetaVault currentVault = IRibbonThetaVault(_vaultAddress);
+
+        // check if vault is registered
+        require(registeredVaults[_vaultAddress], "Vault is not registered");
+
+        address currentUser = address(msg.sender);
```
0xTomoyo, line 185: Can remove the `address(`, also can probably just use `msg.sender` instead of `currentUser`

contracts/vaults/VaultPauser/RibbonVaultPauser.sol
```diff
@@ -0,0 +1,249 @@
...
+
+        // delete position once transfer (revert to zero)
+        delete pausedPositions[_vaultAddress][currentUser];
+
+        // stETH transfers suffer from an off-by-1 error
+        // since we received STETH , we shall deposit using STETH instead of ETH
+        if (_vaultAddress == STETH_VAULT) {
+            totalWithdrawAmount = totalWithdrawAmount - 3;
+
+            emit Resume(currentUser, _vaultAddress, totalWithdrawAmount - 1);
```
0xTomoyo, line 212: Whats the reasoning for emitting `totalWithdrawAmount - 1` rather than just `totalWithdrawAmount`

0xTomoyo (dismissed):

contracts/vaults/VaultPauser/RibbonVaultPauser.sol
kenchangh: good question, not sure, @StephKua if you don't mind chiming in

StephKua: It was because we are going to make another transfer after that, that transfer will ended up sending totalWithdrawAmount - 1, so to align with the true transfer amount, i manually - 1 when emitting the event

test/RibbonThetaSTETHVault.ts
```diff
@@ -3335,6 +3335,44 @@ function behavesLikeRibbonOptionsVault(params: {
       });
     });
 
+    describe("#pauserOwnerAndKeeper", () => {
```
kenchangh, line 3338: Each `describe('#` block refers to a function we are testing, so we should put these cases under the `#pausePosition`. Example:

```js
describe('#pausePosition', () => {
  describe('pause owner and keeper', () => {
    it("returns the owner", async function () {
  })
})
```

kenchangh (approved):

PR #257, opened 2022-05-26
feat: allow delegates

stevenwal:
Added a feature where a wallet can assigned an authorized delegate. This is for multisigs.


kenchangh (approved):
LGTM

PR #271, opened 2022-07-07
feat: upgradable swap contract

stevenwal:
- [x] Make Swap Contract upgradable
- [x] Support swap contract on stETH vault


contracts/vaults/STETHVaultWithSwap/base/RibbonVault.sol
```diff
@@ -0,0 +1,795 @@
+// SPDX-License-Identifier: MIT
+pragma solidity =0.8.4;
```
kenchangh, line 2: My suggestion would be to just implement the changes into RibbonSTETHVault directly, rather than add a new RibbonThetaSTETHVaultWithSwap. This is because we won't be regressing back into Gnosis auctions anymore and it's easier to view the diff

kenchangh: Same for the Yearn contracts

stevenwal: Do we need to do it for yearn contracts? Since we will be reverting YV to base theta vault

contracts/storage/RibbonThetaVaultStorage.sol
```diff
@@ -53,8 +53,8 @@ abstract contract RibbonThetaVaultStorageV7 {
 }
 
 abstract contract RibbonThetaVaultStorageV8 {
-    // Executor role for Swap offers
-    address public offerExecutor;
+    // DEPRECATED: Executor role for Swap offers
+    address private offerExecutor;
```
kenchangh, line 57: could you help prefix this with an underscore too?

kenchangh (dismissed):

kenchangh (approved):

PR #277, opened 2022-07-12
Add Spell Vault

chuddster:


test/RibbonTreasuryVault.ts
```diff
@@ -578,7 +618,7 @@ function behavesLikeRibbonOptionsVault(params: {
         secondOptionExpiry = moment(latestTimestamp * 1000)
           .endOf("month")
           .add(chainId === CHAINID.AVAX_MAINNET ? 0 : 1, "weeks")
-          .add(1, "month")
+          .add(asset === SPELL_ADDRESS[chainId] ? 0 : 1, "month")
```
chuddster, line 621: since block timestamp we use for SPELL is awkward date in month (July 14th), our math for mid-month gets second option expiry wrong - this is workaround

test/RibbonTreasuryVault.ts
```diff
@@ -3262,6 +3308,10 @@ function behavesLikeRibbonOptionsVault(params: {
           .to.emit(vault, "OpenShort")
           .withArgs(secondOptionAddress, depositAmount.mul(2), keeper);
 
+        if (asset === SPELL_ADDRESS[chainId]) {
```
chuddster, line 3311: similar reason - because of expiry of second option we have to multiply premium

kenchangh (approved):

PR #286, opened 2022-08-29
Paradigm fees

kenchangh:


contracts/utils/Swap.sol
```diff
@@ -354,6 +361,14 @@ contract Swap is
             errCount++;
         }
 
+        if (
```
stevenwal, line 364: We can move this validation to `createOffer` so it doesn't confuse bidders since they can't take any action to solve this error. We can do the same for the `sellerAllowance` and `sellerBalance`.

stevenwal (commented):
The rest looks good

stevenwal (dismissed):

stevenwal (approved):

PR #293, opened 2022-11-29
Sahil/airdao

Sahilgat:


contracts/libraries/VaultLifecycleTreasuryBare.sol
```diff
@@ -0,0 +1,827 @@
...
+        }
+
+        bool isPut = vaultParams.isPut;
+        address underlying = vaultParams.underlying;
+        address asset = vaultParams.asset;
+
+        (strikePrice, delta) = closeParams.lastStrikeOverrideRound ==
+            vaultState.round
+            ? (closeParams.overriddenStrikePrice, 0)
+            : IStrikeSelection(strikeSelection).getStrikePrice(expiry, isPut);
```
Sahilgat, line 116: Did we shift it here for code cleanliness compared to #277?

scripts/deploy/28_samb_treasury_vault.ts
```diff
@@ -0,0 +1,100 @@
...
+      },
+    }
+  );
+
+  const initArgs = [
+    {
+      _owner: owner,
+      _keeper: keeper,
+      _feeRecipient: feeRecipient,
+      _managementFee: 0,
```
Sahilgat, line 54: Compared to the original SPELL pr #277, we now seem to use management fee 0 across the vaults. Confirming this is intended behavior.

contracts/libraries/VaultLifecycleTreasuryBare.sol
```diff
@@ -0,0 +1,827 @@
...
+            underlying,
+            asset,
+            strikePrice,
+            expiry,
+            isPut
+        );
+
+        // get the black scholes premium of the option
+        premium = optionsPremiumPricer == address(0)
+            ? 1
```
Sahilgat, line 133: Why do we set it to 1? Is this a bypass in the case where there is no pricer in which case we don't want to revert the function call and instead leave it up to the caller to calculate premiums?

scripts/utils/constants.ts
```diff
@@ -11,6 +11,7 @@ export const STRIKE_STEP = {
   BADGER: 0.1 * 1e8,
   BAL: 0.1 * 1e8,
   SPELL: 1000, // 0.00001 * 1e8
+  SAMB: 1000, // 0.00001 * 1e8
```
Sahilgat, line 14: Will we be using this strike step? I just copied SPELL's for now, but more generally how do we choose the strike step to use?

constants/constants.ts
```diff
@@ -151,6 +151,14 @@ export const SPELL_OWNER_ADDRESS = {
   [CHAINID.ETH_MAINNET]: "0x2ee555c9006a9dc4674f01e0d4dfc58e013708f0",
 };
 
+export const SAMB_ADDRESS = {
+  [CHAINID.ETH_MAINNET]: "0x278a8c33Ef2A7D7C682f458C85E9447ce3645188",
+};
+
+export const SAMB_OWNER_ADDRESS = {
+  [CHAINID.ETH_MAINNET]: "0x6d5b28659b6e76c5843c963f44158c3be6ede3a1",
```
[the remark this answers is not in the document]
chuddster: SAMB_OWNER_ADDRESS is a misnomer - its used as an address that holds a significant amount of the token so that for testing we can simulate transfers from that address to our own test addresses to play around with the token as we see fit

contracts/libraries/VaultLifecycleTreasuryBare.sol
```diff
@@ -0,0 +1,827 @@
...
+            vaultParams,
+            underlying,
+            asset,
+            strikePrice,
+            expiry,
+            isPut
+        );
+
+        // get the black scholes premium of the option
+        premium = optionsPremiumPricer == address(0)
```
chuddster, line 132: can we acctually do `== address(1)`

contracts/libraries/VaultLifecycleTreasuryBare.sol
chuddster: correct. we calculate premiums just as a guideline for the auctions but its not _necessary_ to have. and since we wont have a options premium pricer for this - in case where we set it to address(1) instead of reverting we will just set it to 1

scripts/deploy/28_samb_treasury_vault.ts
```diff
@@ -0,0 +1,100 @@
...
+    contract: "VaultLifecycleTreasuryBare",
+    from: deployer,
+  });
+  console.log(`VaultLifeCycleTreasuryBare @ ${lifecycle.address}`);
+
+  const RibbonTreasuryVault = await ethers.getContractFactory(
+    "RibbonTreasuryVault",
+    {
+      libraries: {
+        VaultLifecycleTreasury: lifecycle.address,
```
chuddster, line 44: lets rename to `VaultLifecycleTreasuryBare: lifecycle.address`

scripts/deploy/28_samb_treasury_vault.ts
chuddster: yep, 0 management fees

scripts/utils/constants.ts
chuddster: lets use 100000, // 0.001 * 1e8

scripts/deploy/28_samb_treasury_vault.ts
Sahilgat: Addressed in [e34465b](https://github.com/ribbon-finance/ribbon-v2/pull/293/commits/e34465b30162c78b1c9313d3468d42cd1c325bc4)

scripts/utils/constants.ts
Sahilgat: Addressed in [e34465b](https://github.com/ribbon-finance/ribbon-v2/pull/293/commits/e34465b30162c78b1c9313d3468d42cd1c325bc4). How did you decide what to use actually?

contracts/libraries/VaultLifecycleTreasuryBare.sol
Sahilgat: Addressed in [e34465b](https://github.com/ribbon-finance/ribbon-v2/pull/293/commits/e34465b30162c78b1c9313d3468d42cd1c325bc4)

scripts/utils/constants.ts
chuddster: based on the price of the asset, we decide a reasonable increment. for example if asset is $100, we might increment in intervals of $1 for optimal granularity of strike price. if asset is $10, we might use $0.1.

contracts/libraries/VaultLifecycleTreasuryBare.sol
chuddster: yeah just better

chuddster (approved):

PR #297, opened 2022-12-21
[WIP] Sahil/add treasury swap

Sahilgat:


contracts/vaults/TreasuryVault/RibbonTreasuryVaultWithAirSwap.sol
```diff
@@ -0,0 +1,1165 @@
...
+        );
+        require(
+            order.signer.token == USDC,
+            "Can only buy with USDC"
+        );
+
+        IAirSwap(AIRSWAP_CONTRACT).swap(order);
+        // oTokens have 8 decimals, get the whole number oTokens sold
+        // oToken premium set will be in USDC and 6 decimals as well
+        uint256 oTokenPremiumToSet = order.signer.amount.div((order.sender.amount.div(10**8)));
```
Sahilgat, line 934: Could I get a sanity check on this calculation? The aim is to have this set the premium per oToken and thus should be 6 decimals

test/RibbonTreasuryVaultWithAirSwap.ts
```diff
@@ -0,0 +1,3679 @@
...
+        totalDistributed = BigNumber.from(bid).sub(performanceFeeInAsset);
+
+        userBalanceAfter = await premiumContract.balanceOf(user);
+        ownerBalanceAfter = await premiumContract.balanceOf(owner);
+
+        assert.bnEqual(
+          userBalanceAfter.sub(userBalanceBefore),
+          BigNumber.from(0)
+        );
+        assert.bnEqual(
```
Sahilgat, line 3325: The test for the BAL vault is failing with an off-by-one error here as shown below:
image
I'm not too sure why this might be the case, since the BAL vault uses the same implementation as the other vaults and has the same number of decimals. I also the original tests in `RibbonTreasuryVault.ts` had some custom logic that excluded the BAL vault, but I'm not sure what was the reason for this.
telegram-cloud-photo-size-5-6147496062563234851-y

contracts/vaults/TreasuryVault/RibbonTreasuryVaultWithAirSwap.sol
Spablob: From what I understood (correct me if I'm wrong):

- `order.signer.amount` has 6 decimals | because there is require above stating it must be USDC which has 6 decimals
- `order.sender.amount` has 8 decimals | because it's the oToken

I would suggest instead:
- `uint256 oTokenPremiumToSet = order.signer.amount.mul(10**8).div((order.sender.amount));`

test/RibbonTreasuryVaultWithAirSwap.ts
Spablob: image
Part 1/3 - Added 6 console.log on the .sol file to check what was going on:

Spablob: image
Part 2/3 - Checked the output of the console.logs - All fine for PERP, BADGER and SPELL:

`console.log("AFTER DISTRIBUTION - stableAsset.balanceOf(address(this))", stableAsset.balanceOf(address(this)));` is correctly showing 0

Spablob: image



Part 3/4 - It's not related to the token BAL itself. It could happen to any token, it just happened to be BAL by chance, due to the number rounding.

Spablob: Part 4/4 - I would suggest following the original testing script - eg:

        assert.bnLte(
          ownerBalanceAfter.sub(ownerBalanceBefore),
          totalDistributed.add(1)
        );
        assert.bnGte(
          ownerBalanceAfter.sub(ownerBalanceBefore),
          totalDistributed.sub(1)
        );

It's also ok to use a similar workaround as was previously used:
image

Spablob:
image

The `setOptionsPremiumPricer` function seems to still exist in `RibbonTreasuryVaultWithAirSwap.sol` but the related tests were removed

contracts/vaults/TreasuryVault/RibbonTreasuryVaultWithAirSwap.sol
```diff
@@ -0,0 +1,1165 @@
...
+
+    event CloseShort(
+        address indexed options,
+        uint256 withdrawAmount,
+        address indexed manager
+    );
+
+    event NewOptionStrikeSelected(uint256 strikePrice, uint256 delta);
+
+    event PremiumDiscountSet(
```
Spablob, line 132: function associated with this event was removed

test/RibbonTreasuryVaultWithAirSwap.ts
```diff
@@ -0,0 +1,3679 @@
...
+        );
+
+        assert.bnEqual(
+          await premiumContract
+            .connect(userSigner)
+            .balanceOf(userSigner.address),
+          BigNumber.from(initialUserSignerBalance).sub(bid)
+        );
+
+        assert.bnEqual(
```
Spablob, line 1590: If checking whether a transfer has successfully occurred I would suggest using a testing pattern similar to:

const BalBefore = ...

const tx = (some tx occurs)

const BalAfter = ...

 assert.bnEqual(BalAfter.sub(BalBefore), amountTransferred)

test/RibbonTreasuryVaultWithAirSwap.ts
```diff
@@ -0,0 +1,3679 @@
...
+
+        await premiumContract
+          .connect(userSigner)
+          .approve(airswapContract.address, bid);
+
+        await vault.connect(keeperSigner).sellOptions(signedOrder);
+
+        // since some oTokens sold, we distribute the premiums
+        await vault.connect(keeperSigner).chargeAndDistribute();
+
```
Spablob, line 1485: image

I'd say we can keep a similar test as in the original file here:

eg.
        assert.bnEqual(
          await otoken.balanceOf(vault.address),
          initialOtokenBalance.div(2)
        );

contracts/vaults/TreasuryVault/RibbonTreasuryVaultWithAirSwap.sol
```diff
@@ -0,0 +1,1165 @@
...
+        uint256 mintAmount =
+            VaultLifecycleTreasuryWithAirSwap.createShort(
+                GAMMA_CONTROLLER,
+                MARGIN_POOL,
+                newOption,
+                lockedBalance
+            );
+
+        IERC20 optionToken = IERC20(newOption);
+        optionToken.safeApprove(address(AIRSWAP_CONTRACT), mintAmount);
```
Spablob, line 906: Since the variable optionToken is only used once - suggestion would be to simplify to 1 line:

`IERC20(newOption).safeApprove(address(AIRSWAP_CONTRACT), mintAmount);`

Sahilgat: Fixed in [a30c9d0](https://github.com/ribbon-finance/ribbon-v2/pull/297/commits/a30c9d05c95b9d301e7c16500455108d96450e59)

contracts/vaults/TreasuryVault/RibbonTreasuryVaultWithAirSwap.sol
Sahilgat: Fixed in [a30c9d0](https://github.com/ribbon-finance/ribbon-v2/pull/297/commits/a30c9d05c95b9d301e7c16500455108d96450e59)

contracts/vaults/TreasuryVault/RibbonTreasuryVaultWithAirSwap.sol
Sahilgat: Fixed in [a30c9d0](https://github.com/ribbon-finance/ribbon-v2/pull/297/commits/a30c9d05c95b9d301e7c16500455108d96450e59)

Sahilgat:
> image
> 
> The `setOptionsPremiumPricer` function seems to still exist in `RibbonTreasuryVaultWithAirSwap.sol` and the related tests were removed. Is the function supposed to be there?

The options premium pricer is used as a parameter in the strike selection contract, and is also used directly in the original  `RibbonTreasuryVault.sol` contract to calculate the minimum bid oToken premium price for Gnosis auctions. Since we are no longer using Gnosis auctions, we can remove the optionsPremiumPricer reference from within the vault contract. Thanks for catching this!

test/RibbonTreasuryVaultWithAirSwap.ts
Sahilgat: Fixed in [a30c9d0](https://github.com/ribbon-finance/ribbon-v2/pull/297/commits/a30c9d05c95b9d301e7c16500455108d96450e59)

test/RibbonTreasuryVaultWithAirSwap.ts
Sahilgat: Addressed in [8928237](https://github.com/ribbon-finance/ribbon-v2/pull/297/commits/892823778c487e6773542058138ce27725910b5a)

test/RibbonTreasuryVaultWithAirSwap.ts
Sahilgat: Addressed in [8928237](https://github.com/ribbon-finance/ribbon-v2/pull/297/commits/892823778c487e6773542058138ce27725910b5a)

test/RibbonTreasuryVaultWithAirSwap.ts
Sahilgat: Addressed in [8928237](https://github.com/ribbon-finance/ribbon-v2/pull/297/commits/892823778c487e6773542058138ce27725910b5a)

test/RibbonTreasuryVaultWithAirSwap.ts
```diff
@@ -1483,6 +1483,11 @@ function behavesLikeRibbonOptionsVault(params: {
         // since some oTokens sold, we distribute the premiums
         await vault.connect(keeperSigner).chargeAndDistribute();
 
+        assert.bnLte(
```
Spablob, line 1486: It could be `assert.bnEqual` vs `assert.bnLte`

Sahilgat: Fixed in [4f81e15](https://github.com/ribbon-finance/ribbon-v2/pull/297/commits/4f81e15a4215f4c7c134e35debe6ab5a3c1b7d65)

PR #298, opened 2023-01-04
Sahil/add uniswap vault

Sahilgat:
We will be launching a retail $UNI (Uniswap’s governance token) covered call vault. [Initial proposal](https://www.notion.so/ribbonfinance/uniswap-cc-pool-4b19cec8bf0740ba9527a911f2b11654)

The following steps were taken with the treasury vault [Spell PR](https://github.com/ribbon-finance/ribbon-v2/pull/277/files) used as template since it is similar.

1. Updated constants
2. Updated deployment script
3. Added test

Next steps:
If deployment script looks good, deploy and upload artefacts in a new PR


test/RibbonThetaVaultWithSwap.ts
```diff
@@ -3581,7 +3612,10 @@ function behavesLikeRibbonOptionsVault(params: {
             .connect(adminSigner)
             .transfer(vault.address, premiumAmount); // Transfer 50 tokens to vault
 
-          await vault.connect(ownerSigner).initiateWithdraw(depositAmount); // User 0 initiates 5000 share withdraw
```
Sahilgat, line 3584: This test `initiated withdraw is completed in a later round` and the following test `vault losses locking up withdraws` were modified as it is buggy. In the original tests, user 1 withdraws his entire share balance from the vault, and the test passes when running for the WBTC vault due to a rounding error that leaves some dust amount that gets rolled over when rolling the vault at the end of round 3. Without this rounding error, this roll fails as the vault has no `lockedAmount` to rollover. The reason this was exposed now is because UNI has 18 decimals while WBTC (the original sole test) has 8 decimals.

In the original `RibbonThetaVault` test without swap which also contained this particular test, it only passed due to this [condition](https://github.com/ribbon-finance/ribbon-v2/blob/master/test/RibbonThetaVault.ts#L3795) which meant that the test was only being run for the WBTC vault again.

Since the test was just meant to verify `pricePerShare` logic, the purpose is still achieved simply by adjusting the amount withdrawn by user 1 to make sure there is still some tokens left in the vault to allow a rollover.

This broken test actually points to a bug in the smart contract. I haven't researched it in-depth at this point since the bug is niche (only occurs if the vault ends up with 0 `lockedAmount` balance), but it likely means some checks are needed either when calling `initiateWithdraw` or `closeRound`. The former updates the `currentQueuedWithdrawShares`, which in turn updates the `lockedBalance` through the latter when `closeRound` is called.

chuddster (approved):

PR #302, opened 2023-01-30
Update sAMB Treasury Vault

chuddster:


contracts/vaults/TreasuryVault/RibbonTreasuryVaultBare.sol
```diff
@@ -0,0 +1,1228 @@
...
+    OwnableUpgradeable
+} from "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+import {
+    ERC20Upgradeable
+} from "@openzeppelin/contracts-upgradeable/token/ERC20/ERC20Upgradeable.sol";
+
+import {Vault} from "../../libraries/Vault.sol";
+import {
+    VaultLifecycleTreasuryBare
+} from "../../libraries/VaultLifecycleTreasuryBare.sol";
```
Sahilgat, line 22: This file isn't uploaded, so we can either remove `RibbonTreasuryVaultBare` or add `VaultLifecycleTreasuryBare`. To prevent confusion for the future and better documentation maybe we can add `VaultLifecycleTreasuryBare` along with a comment in the contract `RibbonTreasuryVaultBare` on when it should be used and how it differs from the original.

scripts/deploy/28_samb_treasury_vault.ts
```diff
@@ -36,7 +36,7 @@ const main = async ({
   console.log(`VaultLifeCycleTreasuryBare @ ${lifecycleAddress}`);
 
   const RibbonTreasuryVault = await ethers.getContractFactory(
-    "RibbonTreasuryVault",
+    "RibbonTreasuryVaultBare",
```
Sahilgat, line 39: Since we also reference `RibbonTreasuryVaultBare` here, if we chose to remove `RibbonTreasuryVaultBare` to prevent similar copies we can add a comment explaining what this is and when/why it should be used.

constants/constants.ts
```diff
@@ -160,11 +160,11 @@ export const SPELL_OWNER_ADDRESS = {
 };
 
 export const SAMB_ADDRESS = {
-  [CHAINID.ETH_MAINNET]: "0x278a8c33Ef2A7D7C682f458C85E9447ce3645188",
+  [CHAINID.ETH_MAINNET]: "0x683aae5cD37AC94943D05C19E9109D5876113562",
 };
 
 export const SAMB_OWNER_ADDRESS = {
-  [CHAINID.ETH_MAINNET]: "0x6d5b28659b6e76c5843c963f44158c3be6ede3a1",
+  [CHAINID.ETH_MAINNET]: "0x39f36935e07f283cf1fd58984c3ed2e63e8b5ccd",
```
Sahilgat, line 167: Since this address is mainly used for testing, I think we can leave it as `0x6d5b28659b6e76c5843c963f44158c3be6ede3a1`. This address has a balance of 10,000 as opposed to the changed `0x39f36935e07f283cf1fd58984c3ed2e63e8b5ccd ` which only has a balance of 1.5

Sahilgat (approved):

PR #305, opened 2023-02-06
vault deployment event emitter

Spablob:


contracts/utils/VaultDeploymentHelper.sol
```diff
@@ -0,0 +1,21 @@
...
+pragma solidity =0.8.4;
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+contract VaultDeploymentHelper is Ownable {
+    /**
+     * @notice Emitted when a new vault is deployed
+     * @param vaultAddress the address of the new vault
+     */
+    event NewVault(address vaultAddress);
```
chuddster, line 11: adjust event for vault type (earn/vip/normal/treasury)

contracts/utils/VaultDeploymentHelper.sol
```diff
@@ -0,0 +1,21 @@
...
+
```
chuddster, line 12: 1. maybe we can also maintain an **array** of all the vault addresses we have and just update it. can be fixed size array + private counter we maintain. 
2. maybe maintain mapping from vault address => enum type

contracts/utils/VaultDeploymentHelper.sol
```diff
@@ -0,0 +1,21 @@
...
+    /**
```
chuddster, line 13: add constructor where we do bulk update of array for all current vaults we have

contracts/utils/VaultDeploymentHelper.sol
Spablob: Addressed

contracts/utils/VaultDeploymentHelper.sol
Spablob: Added mapping enum type => array of all vaults with that type

Currently, the contract has been renamed to `VaultDeploymentEventEmitter`. Should we rename it to "AddressBook" or keep it as it is?

contracts/utils/VaultDeploymentHelper.sol
Spablob: Addressed

scripts/deploy/20_wbtc_call_theta_vault_swap.ts
```diff
@@ -31,6 +31,7 @@ const main = async ({
   );
 
   const manualVolOracle = await deployments.get("ManualVolOracle");
+  const vaultDeploymentEventEmitter = await deployments.get("VaultDeploymentEventEmitter");
```
Spablob, line 34: deployment pending

Spablob: deployment completed

contracts/utils/VaultDeploymentHelper.sol
chuddster: Lets change to AddressBook since it serves that purpose as well - emitter can be secondary

chuddster (approved):

PR #313, opened 2023-06-06
revert autocall buyer changes and deploy

Spablob:


scripts/deploy/32_eth_autocall_vault.ts
```diff
@@ -0,0 +1,209 @@
...
+      USDC_ADDRESS[chainId],
+      OTOKEN_FACTORY[chainId],
+      GAMMA_CONTROLLER[chainId],
+      MARGIN_POOL[chainId],
+    ],
+    libraries: {
+      VaultLifecycleTreasury: lifecycleTreasury.address,
+    },
+  });
+
```
chuddster, line 104: can use same options premium pricer we use for put-selling vault
can use new manual strike selection + lifecycle treasury

scripts/deploy/32_eth_autocall_vault.ts
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  // Autocall vault specific initArgs
+  const initOptionType = 0;
+  const initCouponType = 3;
+  const initAB = 10500;
+  const initNAB = 10500;
+  const initCB = 10500;
+  const initNCB = 10500;
+  const obsFreq = 518400; // 6 days
```
chuddster, line 137: can set initNAB / initNCB to 0

test/RibbonAutocallVault.ts
```diff
@@ -1999,27 +1936,27 @@ function behavesLikeRibbonOptionsVault(params: {
...
         );
 
-        // user transfers the deposit amount and receives 3 coupons earned
+        // user transfers the deposit amount
         const earnedAmt = premiumAmount.div(numTotalObs).mul(3);
         assert.equal(
           userUSDCBalBefore.sub(userUSDCBalAfter).toString(),
-          depositAmount.sub(earnedAmt).toString()
+          depositAmount.toString()
         );
```
chuddster, line 1945: can we check that price per share increases?

scripts/deploy/32_eth_autocall_vault.ts
Spablob: https://github.com/ribbon-finance/ribbon-v2/pull/313/commits/c9870e9f6646b06c8e618677cca57b82b347312c addressed the above with a small fix.

made a small fix in the .sol `initialize` function to ensure that state variables AB and CB are checked instead of nAB and nCB since the former are the ones that will be used on l120 and l121 of the .sol file.

scripts/deploy/32_eth_autocall_vault.ts
Spablob: https://github.com/ribbon-finance/ribbon-v2/pull/313/commits/7cf6f18f834bf6281f3be92a03e395c7a5038916

test/RibbonAutocallVault.ts
Spablob: there is a check on `accountVaultBalance` on these tests which is correctly increasing and uses the price per share in the calculation. seemed more complete given that it checks that both pps is increasing and that is increasing by the correct amount. In any case happy to add a more explicit check with just pps

PS: pps behavior by design may not be linearly up over the trade cycle - eg:

t0: vault is empty
t1: user deposits -> eg. pps = 100
t2: premium is paid mm to vault -> pps = 125
t3: call commitAndClose with 40% premium returning to MM -> pps = 115

so it does increase from t1->t3 (but if there are any returned coupons - this case 40% were returned - pps can possibly reduce from t2->t3)

scripts/deploy/32_eth_autocall_vault.ts
```diff
@@ -0,0 +1,177 @@
...
+      _feeRecipient: feeRecipient,
+      _managementFee: 0,
+      _performanceFee: 0,
+      _tokenName: "ETH Autocall Vault",
+      _tokenSymbol: "ETH-AUTO",
+      _optionsPremiumPricer: pricer.address,
+      _strikeSelection: strikeSelection.address,
+      _premiumDiscount: PREMIUM_DISCOUNT,
+      _auctionDuration: AUCTION_DURATION, // arbitrary value since it is not used in any case
+      _period: 14, // 14 days
```
chuddster, line 121: so a 2 week expiry?

scripts/deploy/32_eth_autocall_vault.ts
```diff
@@ -0,0 +1,177 @@
...
+        OTOKEN_FACTORY[chainId],
+        GAMMA_CONTROLLER[chainId],
+        MARGIN_POOL[chainId],
+      ],
+    });
+  } catch (error) {
+    console.log(error);
+  }
+
+  // Autocall vault specific initArgs
```
chuddster, line 100: lets move these to constants.ts

scripts/deploy/32_eth_autocall_vault.ts
Spablob: confirmed

scripts/deploy/32_eth_autocall_vault.ts
Spablob: https://github.com/ribbon-finance/ribbon-v2/pull/313/commits/62a15070dad5583bac0d30ac5f96e41c55532941

contracts/vaults/TreasuryVault/RibbonAutocallVault.sol
```diff
@@ -123,6 +123,7 @@ contract RibbonAutocallVault is RibbonTreasuryVaultLite, AutocallVaultStorage {
         nObsFreq = _obsFreq;
         nPeriod = period;
         autocallSeller = _autocallSeller;
```
chuddster, line 125: can remove `autocallSeller = _autocallSeller;` now

Spablob: it was decided to be keep this line due to the possible "ERC20: transfer to the zero address" error occurring when rolling to next round on the first round | in the case someone made a USDC transfer to the contract

chudnov (approved):

